PR #103, opened 2023-07-09
Give example CLI opts

jozuas:
One line change to clarify what options are expected to be specified.


mhanberg:
Thanks for the contribution!

I made a different change in another PR, so tho, so I'll be closing this.

PR #112, opened 2023-07-14
Reference lookup perf improvement and table schema versioning

sineed:
depends on https://github.com/elixir-tools/gen_lsp/pull/37

The fix was simple to replace `:dets.select/2` in favor of `:dets.lookup/2`. But it required a change in how references are stored. So the mechanism to clear dets table and trigger recompilation was added. It should recompile initially as the current table schema version is `nil` and every time the schema version value (for example, `@reference_table_schema_version_value`) is changed. I covered both symbol and reference tables.

This will also conflict with the hovering PR as I moved symbol logic to its own GenServer there but it should be not a big deal to resolve here or there

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ next_ls-*.tar
 
 # Temporary files, for example, from tests.
 /tmp/
+
+/.elixir-tools/
+.DS_Store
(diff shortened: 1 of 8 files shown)


mhanberg:
I switched the symbol/ref tables to use sqlite3 instead, so now lookups are very fast.

PR #113, opened 2023-07-17
fix: add .gitignore file to .elixir-tools directory

### LINKED ISSUE #118: suggestion: add .gitignore `*` to .next-ls dir
Elixir LS adds a `.gitignore` with an `*` as the only content, which removes the need for each of us to add `/.next-ls` to our `.gitignore`.

lud-wj:
Fixes #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,10 @@ Using eglot:
 
 ## Installation
 
-The preferred way to use NextLS is through one of the supported editor extensions.
+The preferred way to use Next LS is through one of the supported editor extensions.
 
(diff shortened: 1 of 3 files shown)


mhanberg:
You need to add `.elixir-tools` to your project's gitignore or your global gitignore.

lud-wj:
Whenever I open an elixir project, and I have a lot, this directory will be created. Requiring users to update all their project's .gitignore files will lead to frustration, as some will inevitably forget it and start to add thousand files to their next commit.

A vscode extension should have no impact on the repository it is used in, in my opinion. This proposal aims to do just that :  the directory ignores itself and everyone is happy.

mhanberg:
That's why you can add it to your global gitignore. 

But if this works, then might as well add it. I'll test it out tonight.

PR #131, opened 2023-07-30
fix: swap out dets for sqlite3

mhanberg:
fix: swap out dets for sqlite3

Also fixes a bug with go to definition. it turns out that multiple
references were being stored that overlapped in range, so we'd get
multiple references, which didn't match the pattern, so it returned nil.

Now, sqlite easily deletes any references that overlap with the one we
are inserting in that moment.


mhanberg:
* main
  * **PR #131** <a href="https://app.graphite.dev/github/pr/elixir-tools/next-ls/131" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/elixir-tools/next-ls/131?utm_source=stack-comment).

PR #139, opened 2023-08-05
feat: find references

### LINKED ISSUE #43: references

crbelaus:
This PR is meant to implement the [Find References Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references).

The strategy is as follows:

1. Identify what kind of symbol we looking for
    1. Check the `symbols` table to see if we are in a definition: `defmodule`, `def`, `defp`, `defmacro` or `defstruct`
    2. If we are not in a definition, check the `references` table to see if we are in a function call or in a module alias.
2. Once we know what kind of symbol we are looking for (either a module or a function), check the `references` table to list all known references.

I've been using this today at $DAYJOB, which is a pretty big codebase, and everything seems to work just fine. I've also added two automated tests to confirm that the functionality works for finding function and module references.

Closes #43

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -173,48 +173,19 @@ defmodule NextLS do
     {:reply, symbols, lsp}
   end
 
-  # TODO extract to a separate module
-  # TODO make this work for modules and structs
-  # TODO how to handle aliases?
-  # TODO can we do this in a single DB query?
   def handle_request(%TextDocumentReferences{params: %{position: position, text_document: %{uri: uri}}}, lsp) do
-    import NextLS.DB.Query
-
     file = URI.parse(uri).path
     line = position.line + 1
     col = position.character + 1
 
     locations =
       dispatch(lsp.assigns.registry, :databases, fn databases ->
-        for {database_pid, _} <- databases do
-          reference_query = ~Q"""
-          SELECT identifier, arity, type, module
-          FROM "references" refs
-          WHERE refs.file = ?
-            AND refs.start_line <= ? AND refs.end_line >= ?
-            AND refs.start_column <= ? AND refs.end_column >= ?
-          ORDER BY refs.id ASC
-          LIMIT 1
-          """
-
-          symbol_query =
-            ~Q"""
-            SELECT module, type, name
-            FROM "symbols" sym
-            WHERE sym.file = ?
-              AND sym.line = ?
-            ORDER BY sym.id ASC
-            LIMIT 1
-            """
-
+        for {database, _} <- databases do
           references =
-            case DB.query(database_pid, reference_query, [file, line, line, col, col]) do
-              # Function definition
-              [["store_definition", 3, "function", "elixir_def"]] ->
-                [[module, "def", function]] = DB.query(database_pid, symbol_query, [file, line])
-
+            case symbol_info(file, line, col, database) do
+              {:function, module, function} ->
                 DB.query(
-                  database_pid,
+                  database,
                   ~Q"""
                   SELECT file, start_line, end_line, start_column, end_column
                   FROM "references" as refs
@@ -225,25 +196,9 @@ defmodule NextLS do
                   [function, "function", module]
                 )
 
-              # Function reference
-              [[function, arity, "function", module]] ->
+              {:module, module} ->
                 DB.query(
-                  database_pid,
-                  ~Q"""
-                  SELECT file, start_line, end_line, start_column, end_column
-                  FROM "references" as refs
-                  WHERE refs.identifier = ?
-                    AND refs.arity = ?
-                    AND refs.type = ?
-                    AND refs.module = ?
-                  """,
-                  [function, arity, "function", module]
-                )
-
-              # Module reference
-              [[_alias, _, "alias", module]] ->
-                DB.query(
-                  database_pid,
+                  database,
                   ~Q"""
                   SELECT file, start_line, end_line, start_column, end_column
                   FROM "references" as refs
@@ -253,38 +208,16 @@ defmodule NextLS do
                   [module, "alias"]
                 )
 
-              [] ->
-                case DB.query(database_pid, symbol_query, [file, line]) do
-                  # Module definition
-                  [[module, "defmodule", _]] ->
-                    DB.query(
-                      database_pid,
-                      ~Q"""
-                      SELECT file, start_line, end_line, start_column, end_column
-                      FROM "references" as refs
-                      WHERE refs.module = ?
-                        and refs.type = ?
-                      """,
-                      [module, "alias"]
-                    )
-
-                  _unknown ->
-                    []
-                end
+              :unknown ->
+                []
             end
 
           for [file, start_line, end_line, start_column, end_column] <- references do
             %Location{
               uri: "file://#{file}",
               range: %Range{
-                start: %Position{
-                  line: start_line - 1,
-                  character: start_column - 1
-                },
-                end: %Position{
-                  line: end_line - 1,
-                  character: end_column - 1
-                }
+                start: %Position{line: start_line - 1, character: start_column - 1},
+                end: %Position{line: end_line - 1, character: end_column - 1}
               }
             }
           end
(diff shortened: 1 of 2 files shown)


mhanberg:
> Extract the "find references" logic to a separate module just like the rest of actions such as NextLS.Definition.

I think it's probably fine to leave it "unextracted" for now. I probably want these in the same module anyway, but will name it something different. Can do it later.

> Make the implementation work for modules and struct. At the moment it only works for functions.

You can do this in a later PR if you want, would be better to get any version of this working and released so that the general infrastructure can be used and tested. (same thing i did with go to def basically)

> See how to handle aliases. For example if we want to find references of DB where we have a previous alias NextLS.DB.

The reference stored in the database should be the expanded alias (I think). You can double check by querying the db with the sqlite3 cli or using a GUI client. So it should "just work" for aliases.

> See if I can get the information in a single query instead of two.

We can optimize this later, feel free to not worry about it for now

lib/next_ls.ex
```diff
@@ -171,6 +173,66 @@ defmodule NextLS do
...
+        for {database_pid, _} <- databases do
+          query = ~Q"""
+          SELECT identifier, arity, type, module
+          FROM "references" as refs
+          WHERE refs.file = ?
+            AND refs.start_line <= ? AND refs.end_line >= ?
+            AND refs.start_column <= ? AND refs.end_column >= ?
+          ORDER BY refs.id ASC
+          LIMIT 1;
+          """
```
mhanberg, line 198: one thing to consider is this should work if you trigger "find references" on a function/module definition _or_ a reference.

Might need to add end_line,end_col to the symbols table, so we can figure out if we're "inside" one.

lib/next_ls.ex
```diff
@@ -171,6 +173,66 @@ defmodule NextLS do
...
+
+          [[identifier, arity, type, module]] = DB.query(database_pid, query, [file, line, line, col, col])
```
mhanberg, line 200: in case you trigger find references on something that has no reference (like whitespace or somewhere weird), you'll want to account for this returning an empty list.

mhanberg:
Feel free to post a screen recording if you get a chance, it's always fun to see it in action 😎

lib/next_ls.ex
```diff
@@ -171,6 +173,66 @@ defmodule NextLS do
...
+
+          query = ~Q"""
+          SELECT file, start_line, end_line, start_column, end_column
+          FROM "references" as refs
+          WHERE refs.identifier = ?
+            AND refs.arity = ?
+            AND refs.type = ?
+            AND refs.module = ?
```
mhanberg, line 208: in the case that we are finding references for a reference (meaning, a usage of the function, not the definition), we'll want to exclude the reference we're on from the result set.

I think at least, would be valuable to see what other LSPs like gopls and rust analyzer do

mhanberg:
I know you are still in progress, but i'll drop a couple of screen recordings of some odd behavior i'm seeing as of now

Unrelated results for private function

https://github.com/elixir-tools/next-ls/assets/5523984/ca55abf2-f0db-492c-9aae-b90a46ffc574

client side exception when finding references of a function definition

https://github.com/elixir-tools/next-ls/assets/5523984/68c087fe-15ed-42a6-874d-09b986063cbb

unrelated results for a remote function

https://github.com/elixir-tools/next-ls/assets/5523984/f8636423-19b2-4bcf-bb92-bb428de606aa

lib/next_ls.ex
crbelaus: Yeah that makes sense. This is what I meant as well when talking about aliases earlier.

I think that we may be able to use `Code.Fragment.surround_context/3` to know what we are hovering when trying to find references: a function call, a module reference, a function definition, a module definition, etc. And then use this information to query the `references` table accordingly.

crbelaus: Looks like we have enough information in the symbols and references table. I made an improvement that now works with function/module definitions and references.

crbelaus:
Thanks for early the feedback @mhanberg 🙇 

I've just pushed another commit that makes the "Find references" functionality work from module/function definitions as well as references (previously it was only working from function references).

I thought that I would need [Code.Fragment.surround_context/3](https://hexdocs.pm/elixir/main/Code.Fragment.html#surround_context/3) but it looks that we have all the information that we need in the `references` and `symbols` tables. The code is _very_ messy at the moment as I am still trying to get the right information out. 

Here are a few videos from VSCode with the latest changes:

<details><summary>Find references from a function call</summary>

https://github.com/elixir-tools/next-ls/assets/2965623/5b99f642-9e44-4f7b-b80a-54c5226b8ef1
</details> 

<details><summary>Find references from a function definition</summary>

https://github.com/elixir-tools/next-ls/assets/2965623/cbc4941d-597c-4d69-9de0-9d0c2d7dd4b9
</details> 

<details><summary>Find references from a module reference</summary>

https://github.com/elixir-tools/next-ls/assets/2965623/dfa1e1e5-a5ce-4817-8ac6-1cc1d90d163b
</details> 

<details><summary>Find references from a module definition</summary>


https://github.com/elixir-tools/next-ls/assets/2965623/647e5923-c01f-4385-9bba-bac7d86b1bc9
</details>

lib/next_ls.ex
crbelaus: This should be handled with the latest commit.

lib/next_ls.ex
crbelaus: ElixirLS doesn't exclude the current reference but we can definitely do it. I'll definitely check what other LSPs do.

mhanberg:
🤯

image

crbelaus:
I've just pushed another commit to extract and improve the symbol information query. We now check if we are in a symbol definition first and fallback to check if we are in a function/module reference otherwise.

This logic has been moved into the `symbol_info/4` function. I've kept this private function in the same file as I am still unsure what would be the right place for it.

I plan to test this tomorrow for the entire day in my $DAYJOB, which is a pretty big codebase. If I don't find any stability issues or obvious problems I will remove the draft status and open this for review if that's okay for you @mhanberg

mhanberg:
@crbelaus you are a legend!

Thank you so much for this, I'll do a thorough review once you mark as ready, but when I looked yesterday I was already pretty happy.

crbelaus:
I've removed the draft state and this is now ready for review. I've also added a few automated tests to ensure that this works well and doesn't break with future changes.

test/next_ls_test.exs
```diff
@@ -888,6 +888,106 @@ defmodule NextLSTest do
...
+                    [
+                      %{
+                        "uri" => ^uri,
+                        "range" => %{
+                          "start" => %{"line" => 3, "character" => 10},
+                          "end" => %{"line" => 3, "character" => 18}
+                        }
+                      }
+                    ],
+                    500
```
mhanberg, line 955: I have the timeout for these globally set to 30 seconds, so you can remove all of these custom timeouts, as it's actually making it shorter 😂

https://github.com/elixir-tools/next-ls/blob/main/test/test_helper.exs#L9

crbelaus: I've just pushed a commit that removes those explicit timeouts.

mhanberg: 🚀

lib/next_ls.ex
```diff
@@ -171,6 +173,62 @@ defmodule NextLS do
...
+  # TODO handle `context: %{includeDeclaration: true}` to include the current symbol definition among
+  # the results.
+  def handle_request(%TextDocumentReferences{params: %{position: position, text_document: %{uri: uri}}}, lsp) do
+    file = URI.parse(uri).path
+    line = position.line + 1
+    col = position.character + 1
+
+    locations =
+      dispatch(lsp.assigns.registry, :databases, fn databases ->
+        for {database, _} <- databases do
```
mhanberg, line 185: If you are going to flatten it below, you might as well use Enum.flat_map here instead of a comprehension

mhanberg:
Just had that one comment, i will test this manually after work and hopefully merge and release before i present at the Denver Elixir Meetup about NextLS, so I can show off your hard work 💪.

I assume you'll be asleep at that time but I'll give you the loudest of shoutouts 👏.

lib/next_ls.ex
crbelaus: You are absolutely right! Much better to do it in a single operation since it doesn't hurt readability either.

I've just pushed a commit that replaces the for comprehension with `Enum.flat_map` as you suggested.

PR #147, opened 2023-08-09
feat(definition): go to dependency

### LINKED ISSUE #91: [definition] dependency definitions

mhanberg:
This enables the tracers when your dependencies are compiling, which adds them to the symbol table.

TODO

- [ ] Make sure that they do not show up in Workspace Symbols
- [ ] see if we need/can show the progress message when the database is finishing a "batch" of updates (message can say "indexing"). It can get bogged down and the go to def won't work until it's done inserting.
- [ ] tests
- [ ] determine feasibility for elixir source files. they won't exist on disk, and since they aren't compiled, we won't have symbols for them, but there _might_ be the possibility of opening the line of code in github in their web browser.

cc @dvic, feel free to take this over if you'd like!

Closes #91

### DIFF
--- a/priv/monkey/_next_ls_private_compiler.ex
+++ b/priv/monkey/_next_ls_private_compiler.ex
@@ -117,6 +117,12 @@ defmodule :_next_ls_private_compiler do
     # keep stdout on this node
     Process.group_leader(self(), Process.whereis(:user))
 
+    tracers = Code.get_compiler_option(:tracers)
+
+    tracers = Enum.uniq([NextLSPrivate.Tracer | tracers])
+
+    Code.put_compiler_option(:tracers, tracers)
+
     Mix.Task.clear()
 
     # load the paths for deps and compile them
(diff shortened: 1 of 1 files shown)


dvic:
> determine feasibility for elixir source files. they won't exist on disk, and since they aren't compiled, we won't have symbols for them, but there might be the possibility of opening the line of code in github in their web browser.

If you use ASDF and build Elixir from source (i.e., use `elixir ref:v1.15.4` instead of `elixir 1.15.4-otp26`) then the source code navigation works just fine in ElixirLS. I guess we can support this for now?  

> cc @dvic, feel free to take this over if you'd like!

sure :) I'll give it a go!

mhanberg:
ElixirLS uses the elixir sense project instead of only elixir core, which gives it other capabilities. 

If we can determine the function/module using the Code.Fragment module and it's path, then should be good.

mhanberg:
- [ ] make sure dependencies don't show up in Find References

mhanberg:
Superceded by #171

PR #159, opened 2023-08-10
Use sh instead of bash for start script

dvic:
`bin/start` was not working for me, it gave:

```
./bin/start: line 16: ${machine,,}_${arch,,}: bad substitution
```

so I ran it through chatgpt 😅, this works on my machine™️

### DIFF
--- a/bin/start
+++ b/bin/start
@@ -1,25 +1,19 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # used for local development
-
 
 set -e
 
 cd "$(dirname "$0")"/.. || exit 1
 
-function main() {
-  local machine
-  local arch
-  local target
-  machine="$(uname)"
-  arch="$(uname -m)"
-  target="${machine,,}_${arch,,}"
+main() {
+	machine="$(uname)"
+	arch="$(uname -m)"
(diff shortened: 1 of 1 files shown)


bin/start
```diff
@@ -1,25 +1,19 @@
...
-  local target
-  machine="$(uname)"
-  arch="$(uname -m)"
-  target="${machine,,}_${arch,,}"
-
-  local
+main() {
+	machine="$(uname)"
+	arch="$(uname -m)"
+	target="$(echo $machine | tr 'A-Z' 'a-z')_$(echo $arch | tr 'A-Z' 'a-z')"
```
mhanberg, line 12: can you cut this down so the only change is the tr vs the esoteric bash substitution?

i don't want the whitespace changes, changing from bash to sh, or removing the local var declarations

bin/start
```diff
@@ -1,25 +1,19 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # used for local development
 
-
```
mhanberg, line 5: this white space change can stay tho

mhanberg:
Sorry, decided to just not use burrito in dev 😅

Closed by https://github.com/elixir-tools/next-ls/pull/161

PR #165, opened 2023-08-11
fix: improve error handling for compiler diagnostics

zachallaun:
Playing around with Next LS this morning and was running up against some issues compiling a project of mine. I think the previous handling of `:badrpc` errors was incorrect, which was hiding certain errors of causing them to bubble up in weird ways (like trying iterate over an atom).

### DIFF
--- a/lib/next_ls/extensions/elixir_extension.ex
+++ b/lib/next_ls/extensions/elixir_extension.ex
@@ -24,7 +24,7 @@ defmodule NextLS.ElixirExtension do
   end
 
   @impl GenServer
-  def handle_info({:compiler, diagnostics}, state) do
+  def handle_info({:compiler, diagnostics}, state) when is_list(diagnostics) do
     DiagnosticCache.clear(state.cache, :elixir)
 
     for d <- diagnostics do
--- a/lib/next_ls/runtime.ex
+++ b/lib/next_ls/runtime.ex
@@ -128,7 +128,7 @@ defmodule NextLS.Runtime do
                 NextLS.Logger.log(logger, "The runtime for #{name} has successfully shutdown.")
 
               reason ->
-                NextLS.Logger.error(logger, "The runtime for #{name} has crashed with reason: #{reason}.")
+                NextLS.Logger.error(logger, "The runtime for #{name} has crashed with reason: #{inspect(reason)}.")
             end
         end
       end)
@@ -144,8 +144,8 @@ defmodule NextLS.Runtime do
           |> Path.join("monkey/_next_ls_private_compiler.ex")
           |> then(&:rpc.call(node, Code, :compile_file, [&1]))
           |> tap(fn
-            {:badrpc, :EXIT, {error, _}} ->
-              NextLS.Logger.error(logger, error)
+            {:badrpc, error} ->
+              NextLS.Logger.error(logger, {:badrpc, error})
 
             _ ->
               :ok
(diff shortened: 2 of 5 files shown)


lib/next_ls/extensions/elixir_extension.ex
```diff
@@ -24,7 +24,7 @@ defmodule NextLS.ElixirExtension do
   end
 
   @impl GenServer
-  def handle_info({:compiler, diagnostics}, state) do
+  def handle_info({:compiler, diagnostics}, state) when is_list(diagnostics) do
```
mhanberg, line 27: you'll need to handle an unhandled messaged, and emit a warning log

lib/next_ls/runtime.ex
```diff
@@ -248,4 +249,10 @@ defmodule NextLS.Runtime do
       true
     end
   end
+
+  defp dispatch_to_extensions(state, message) do
+    Registry.dispatch(state.registry, :extensions, fn entries ->
+      for {pid, _} <- entries, do: send(pid, message)
+    end)
+  end
```
mhanberg, line 257: i prefer to not extract private functions for things that are only used once

zachallaun: Oops, agreed -- I was using it twice because I thought it was needed elsewhere but forgot to un-extract it after 😊

lib/next_ls/extensions/elixir_extension.ex
zachallaun: Just pushed some changes/fixes for this branch. At this point, I don't think we need to handle unexpected messages here because this *should* always be a list. Prior to these changes, if there was an RPC call failure, this might be like a :nodedown message or something, which was then causing a more confusing downstream error.

At this point, this guard is basically just a safety net so that, if an invalid message is sent, it's more obvious/clear in the future.

zachallaun: Forgot to add: I'm fine either removing this guard or adding a catch-all for unhandled messages, but I personally think it's okay to be assertive here.

mhanberg:
Looks alright, could you add a test or two?

mhanberg:
just a test in the elixir_extension_test.exs is fine, would be too hard to simulate in an integration tests (the tests that use the GenLSP.Test helpers)

lib/next_ls/runtime.ex
```diff
@@ -120,7 +120,7 @@ defmodule NextLS.Runtime do
                 NextLS.Logger.log(logger, "The runtime for #{name} has successfully shutdown.")
 
               reason ->
-                NextLS.Logger.error(logger, "The runtime for #{name} has crashed with reason: #{reason}.")
+                NextLS.Logger.error(logger, "The runtime for #{name} has crashed with reason: #{inspect(reason)}.")
```
zachallaun, line 123: This fixes a `String.Chars` protocol error when the reason is a tuple, for example.

lib/next_ls/runtime.ex
```diff
@@ -136,8 +136,8 @@ defmodule NextLS.Runtime do
           |> Path.join("monkey/_next_ls_private_compiler.ex")
           |> then(&:rpc.call(node, Code, :compile_file, [&1]))
           |> tap(fn
-            {:badrpc, :EXIT, {error, _}} ->
```
zachallaun, line 139: Looking at the `:rpc` docs, I don't think this would ever match, since it should be a 2-tuple. [Docs](https://www.erlang.org/doc/man/rpc#call-4)

lib/next_ls/runtime.ex
```diff
@@ -227,6 +223,17 @@ defmodule NextLS.Runtime do
     {:noreply, %{state | compiler_ref: nil}}
   end
 
+  def handle_info({:DOWN, _, :port, port, reason}, %{port: port} = state) do
+    error = {:port_down, reason}
+    state.on_initialized.({:error, error})
```
mhanberg, line 228: i think that the port could go down after the runtime has initialized, so calling that function here seems not intuitive.

if you say "when not is_map_key(state, :node)" as a clause that implies the node hasn't connected yet

or else it might be better to add another callback to pass in maybe

lib/next_ls/runtime.ex
```diff
@@ -192,15 +188,22 @@ defmodule NextLS.Runtime do
   def handle_call(:compile, from, %{node: node} = state) do
     task =
       Task.Supervisor.async_nolink(state.task_supervisor, fn ->
-        {_, errors} = :rpc.call(node, :_next_ls_private_compiler, :compile, [])
+        case :rpc.call(node, :_next_ls_private_compiler, :compile, []) do
+          {:badrpc, error} ->
+            NextLS.Logger.error(state.logger, "Bad RPC call: #{inspect(error)}")
```
mhanberg, line 193:
```suggestion
            NextLS.Logger.error(state.logger, "Bad RPC call to node #{node}: #{inspect(error)}")
```

PR #171, opened 2023-08-11
feat(definition): go to dependency

dvic:
TODO

- [x] make sure that they do not show up in Workspace Symbols
- [x] make sure dependencies don't show up in Find References
- [x] ~see if we need/can show the progress message when the database is finishing a "batch" of updates (message can say "indexing"). It can get bogged down and the go to def won't work until it's done inserting.~ Done in #179 
- [x] tests
- [ ] determine feasibility for elixir source files. they won't exist on disk, and since they aren't compiled, we won't have symbols for them, but there _might_ be the possibility of opening the line of code in github in their web browser.

https://github.com/elixir-tools/next-ls/assets/1214337/f49547be-efcd-4a61-8a5a-974dc285a508


mhanberg:
tests pass for me locally

mhanberg:
> make sure dependencies don't show up in Find References

I don't think the existing tests handle this, so I wouldn't call it completed yet. same for workspace symbols

test/next_ls_test.exs
```diff
@@ -235,6 +233,9 @@ defmodule NextLSTest do
       assert_result 2, nil
     end
 
+    # dependency should not show up in workspace symbols
+    @tag :with_dependency
```
mhanberg, line 237: I think it is more flexible to make the tag take the dependencies itself

```elixir
@tag dependencies: """
{:jason, "~> 1.4"},
{:foo, "~> 2.0"}
"""
```

then in the mix_exs function, you just interpolate that into the string

I would also unpack the new `write_project` function, as I sense this going to be a brittle function that will churn often.

test/support/utils.ex
```diff
@@ -4,7 +4,37 @@ defmodule NextLS.Support.Utils do
...
+    if Map.get(ctx, :with_dependency) do
+      ext = Path.join(proj_dir, "lib/ext.ex")
+
+      File.write!(ext, """
+      defmodule Ext do
+        def run() do
+          Jason.encode!(%{})
+        end
+      end
+      """)
```
mhanberg, line 24: this should just go in the setup for the relevant describe or test

test/support/utils.ex
```diff
@@ -4,7 +4,37 @@ defmodule NextLS.Support.Utils do
...
+     
+     {msg, exit_code} = System.cmd("mix", ["deps.get"], cd: proj_dir, into: "")
```
mhanberg, line 26: this can probably be in it's own function of `mix_deps_get` and then can be called by the tests as `setup :mix_deps_get`

mhanberg: including the assert below this line

mhanberg:
I think that this section of functionality might warrant it's own test module, as it will be testing workspace symbols, references, and definition.

that way you can restrict the proliferation of code that has to install a dependency to as few tests as possible, as literally downloading a dep  during every test run is bound to be brittle and time consuming.

mhanberg:
actually, it might make more sense to use a path dependency for these tests, and write those files in the same was as we do the normal tests.

actually running mix deps.get inside the tests seems like a red flag to me

dvic:
> actually, it might make more sense to use a path dependency for these tests, and write those files in the same was as we do the normal tests.
> 
> actually running mix deps.get inside the tests seems like a red flag to me

i think you’re right, but this got me thinking: don’t we want go to definition for path dependencies? does this include umbrella applications? how does this relate to workspaces? Let’s say I open a mono repo, it would be nice to have go to definition across projects.

mhanberg:
Path deps should work the same as remote deps. 

Umbrella apps should work already. 

Monorepo project would most likely be a dep

dvic:
@mhanberg what do you want to do actually for the workspace symbols: filter them in the query or avoid inserting them in the DB in te first place? If it's the latter: we do this in the sidecar then?

dvic:
I pushed the former, let me know if you want it differently :)

lib/next_ls/db.ex
```diff
@@ -13,8 +13,8 @@ defmodule NextLS.DB do
   @spec query(pid(), query(), list()) :: list()
   def query(server, query, args \\ []), do: GenServer.call(server, {:query, query, args}, :infinity)
 
-  @spec symbols(pid()) :: list(map())
-  def symbols(server), do: GenServer.call(server, :symbols, :infinity)
+  @spec symbols(pid(), [String.t()]) :: list(map())
+  def symbols(server, workspace_folders), do: GenServer.call(server, {:symbols, workspace_folders}, :infinity)
```
mhanberg, line 17: this function is actually a residual from the pre-sqlite days, and I think we can "unpack" it now. (passing in the workspace folders to me is leaking implementation details and is a sign we want to yield full control back to the caller)

so i would delete this function and associated callback, and move the query it calls to the callsite, and use the query function instead.

test/next_ls/dependency_test.exs
```diff
@@ -0,0 +1,141 @@
...
+        "end" => %{
+          "line" => 1,
+          "character" => 0
+        }
+      },
+      "uri" => ^uri
+    }
+  end
+
+  test "does not show in workspace symbols", context do
```
mhanberg, line 84: still need a similar test for references

dvic: Just to be sure, what kind of test did you have in mind here? Parent calls into dependency and the somewhere in the dependency this same function is called? Would this show up as a function reference? And what about module?

mhanberg: If you do "find references" on something like Enum.map, it'll show locations in the dependencies where it's used. 

Basically when you do "find references" on anything, it should only show you references that occur inside _your_ codebase.

dvic: Done!

lib/next_ls.ex
```diff
@@ -716,4 +716,43 @@ defmodule NextLS do
   end
 
   defp clamp(line), do: max(line, 0)
+
+  defp fetch_symbols(pid, lsp) do
```
mhanberg, line 720: i prefer to not extract functions if they're only going to be used one

mhanberg:
This PR is blocked on several things I am working on. 

1. Improved schema migration strategy 
2. Symbol indexing tracker 
3. Add new column to symbols table

mhanberg:
You should be able to rebase now and work on this again.

There is a new "source" column on the symbols table that defaults to "user".

Then, when you create a separate tracer for compiling the dependencies, you can have that one insert the symbols with the source being set to "dep"

then when you are filtering on them for workspace symbols, you can say `WHERE source = 'user'`

priv/monkey/_next_ls_private_compiler.ex
```diff
@@ -1,5 +1,127 @@
...
+         module: env.module,
+         source: @source
+       }},
+      []
+    )
+
+    :ok
+  end
+
+  def trace({:on_module, bytecode, _}, env) do
```
mhanberg, line 80: i believe all we need is this clause. we don't need to trace references inside dependencies, at least yet.

mhanberg:
https://github.com/elixir-tools/next-ls/assets/5523984/8d126309-ce68-4a3f-be69-09780266ded7

need to figure out how to handle this

mhanberg:
> CleanShot.2023-08-14.at.18.29.04.mp4
> 
> need to figure out how to handle this

I have a fix for this, just writing a test right now

PR #180, opened 2023-08-13
refactor: migration system

mhanberg:
1. Create the `schema` table if needed. This includes a version column.
  2. If the max version selected from the `schema` table is equal to the current version, we noop and halt.
     Else, if the max version is less than the current one (compiled into this module) or nil
     we "upgrade" the database.
  3. Unless halting, we drop the non-meta tables, and then create them from scratch
  4. Return a value to signal to the caller that re-indexing is necessary.

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -133,7 +133,7 @@ defmodule NextLS do
             [] ->
               nil
 
-            [[_pk, _mod, file, _type, _name, line, column] | _] ->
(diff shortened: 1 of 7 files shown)


mhanberg:
* main
  * **PR #180** <a href="https://app.graphite.dev/github/pr/elixir-tools/next-ls/180" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/elixir-tools/next-ls/180?utm_source=stack-comment).

PR #182, opened 2023-08-14
fix: coalesce nil start line to 1

### LINKED ISSUE #160: [NextLS] sqlite3 error: NOT NULL constraint failed: references.start_line
I've installed elixir-tools.vscode from the marketplace, and it's working, but there are tons of errors in the NextLS output.

```
statement: INSERT INTO 'references' (identifier, arity, file, type, module, start_line, start_column, end_line, end_column)
    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?);
[Error - 10:12:27 AM] [NextLS] sqlite3 error: NOT NULL constraint failed: references.start_line
```

elixir-tools: 0.6.0
elixir: 1.15.4-otp-26
erlang: 26.0.2
OS: MacOS Ventura (13.4.1)
VSCode: 1.81.0

dvic:
Fixes #160

### DIFF
--- a/lib/next_ls/db.ex
+++ b/lib/next_ls/db.ex
@@ -148,11 +148,11 @@ defmodule NextLS.DB do
       module: module
     } = reference
 
+    line = meta[:line] || 1
     col = meta[:column] || 0
 
-    {{start_line, start_column}, {end_line, end_column}} =
-      {{meta[:line], col},
-       {meta[:line], col + String.length(identifier |> to_string() |> String.replace("Elixir.", ""))}}
+    {start_line, start_column} = {line, col}
+    {end_line, end_column} = {line, col + String.length(identifier |> to_string() |> String.replace("Elixir.", ""))}
 
     __query__(
       {conn, s.logger},


lib/next_ls/runtime/sidecar.ex
```diff
@@ -15,14 +15,23 @@ defmodule NextLS.Runtime.Sidecar do
   end
 
   def handle_info({:tracer, payload}, state) do
+    payload = replace_nil(payload, :start_line, 1)
```
mhanberg, line 18: i think we just need it on the references, not symbols

lib/next_ls/runtime/sidecar.ex
```diff
@@ -15,14 +15,23 @@ defmodule NextLS.Runtime.Sidecar do
...
     DB.insert_symbol(state.db, payload)
 
     {:noreply, state}
   end
 
   def handle_info({{:tracer, :reference}, payload}, state) do
+    payload = replace_nil(payload, :start_line, 1)
```
mhanberg, line 25: i think we can change the insert query to have `COALESCE(?, 0)` instead of this logic here.

also i think it might be start_line and end_line, but i'll double check

lib/next_ls/runtime/sidecar.ex
dvic: Yup (checked it locally).

lib/next_ls/runtime/sidecar.ex
dvic: When I checked this fix locally, it did not work, because the meta simply does not contain `line`. So I moved the logic to the DB. Is that ok? Or do you want a `Keyword.put_new` in the sidecar?

mhanberg: db

dvic: done

mhanberg: sorry, i meant use the `COALESCE` sql function in the insert query

PR #185, opened 2023-08-15
ci: add build cache to dialyzer

dvic:
I noticed `mix dialyzer` takes >= 1 minute, of which only 4 seconds is dialyzer. This PR should ensure that the built dependencies are also cached.


dvic:
dialyzer job now takes ~22 secs :)

PR #187, opened 2023-08-16
Don't clean references on every file save

### LINKED ISSUE #154: References: determine best place to wipe out a file's references
## Description

If you save a file without actually changing the contents, the compiler won't compile the file again, but we are currently deleting all references from that file, in preparation for them to be replenished.

We should either delete the references when we know we are actually compiling, or conditionally delete them if we can determine if the file is actually dirty since the last save.

crbelaus:
This pull request stops cleaning file references on every file save and, instead, cleans them only when we compile the file.

On the one hand we stop cleaning the file references on the `TextDocumentDidSave` notification. On the other hand we now clean the references on the `:start` event of the compiler tracer.

As per [the Elixir docs](https://hexdocs.pm/elixir/Code.html#module-compilation-tracers):

> `:start` - (since v1.11.0) invoked whenever the compiler starts to trace a new lexical context. A lexical context is started when compiling a new file or when defining a module within a function

Since this is only executed when compiling a file or defining a module we know that it is safe to to clean references at this point, since new ones will be populated during the compilation that just started.

The following video shows an example. Observe how the terminal at the bottom of the screen says "CLEANING REFERENCES FOR $FILE" when I update the file contents but doesn't do anything when I save without changing the file contents.

https://github.com/elixir-tools/next-ls/assets/2965623/476bf62a-5114-4a7b-8772-23fbeea65c88

Closes #154

### DIFF
--- a/lib/next_ls/db.ex
+++ b/lib/next_ls/db.ex
@@ -144,8 +144,6 @@ defmodule NextLS.DB do
     {:message_queue_len, count} = Process.info(self(), :message_queue_len)
     NextLS.DB.Activity.update(s.activity, count)
 
-    NextLS.Logger.warning(s.logger, "CLEANING REFERENCES FOR #{filename}")
-
     __query__(
       {conn, s.logger},
       ~Q"""
--- a/lib/next_ls/db.ex
+++ b/lib/next_ls/db.ex
@@ -19,6 +19,9 @@ defmodule NextLS.DB do
   @spec insert_reference(pid(), map()) :: :ok
   def insert_reference(server, payload), do: GenServer.cast(server, {:insert_reference, payload})
 
+  @spec insert_reference(pid(), String.t()) :: :ok
+  def clean_references(server, filename), do: GenServer.cast(server, {:clean_references, filename})
+
   def init(args) do
     file = Keyword.fetch!(args, :file)
     registry = Keyword.fetch!(args, :registry)
@@ -132,6 +135,24 @@ defmodule NextLS.DB do
           VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?);
       """,
       [identifier, reference[:arity], file, type, module, start_line, start_column, end_line, end_column, source]
+    )
+
+    {:noreply, s}
+  end
+
+  def handle_cast({:clean_references, filename}, %{conn: conn} = s) do
+    {:message_queue_len, count} = Process.info(self(), :message_queue_len)
+    NextLS.DB.Activity.update(s.activity, count)
+
+    NextLS.Logger.warning(s.logger, "CLEANING REFERENCES FOR #{filename}")
+
+    __query__(
+      {conn, s.logger},
+      ~Q"""
+      DELETE FROM 'references'
+      WHERE file = ?;
+      """,
+      [filename]
     )
 
     {:noreply, s}
--- a/lib/next_ls/db.ex
+++ b/lib/next_ls/db.ex
@@ -19,7 +19,7 @@ defmodule NextLS.DB do
   @spec insert_reference(pid(), map()) :: :ok
   def insert_reference(server, payload), do: GenServer.cast(server, {:insert_reference, payload})
 
-  @spec insert_reference(pid(), String.t()) :: :ok
+  @spec clean_references(pid(), String.t()) :: :ok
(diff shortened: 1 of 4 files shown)


lib/next_ls/db.ex
```diff
@@ -137,6 +140,24 @@ defmodule NextLS.DB do
     {:noreply, s}
   end
 
+  def handle_cast({:clean_references, filename}, %{conn: conn} = s) do
+    {:message_queue_len, count} = Process.info(self(), :message_queue_len)
+    NextLS.DB.Activity.update(s.activity, count)
+
+    NextLS.Logger.warning(s.logger, "CLEANING REFERENCES FOR #{filename}")
```
crbelaus, line 147: I added this only for the video, can be removed before merging.

mhanberg:
One thing that I am unsure about with the `:start` trace is whether it is called only for a file, or if it is called multiple times if you define submodules or multiple modules in the same file

so, it should still work even if you have this code

```elixir
# lib/foo.ex
defmodule Foo do
  defmodule Bar do
    def run() do
    end
  end

  def run() do
  end
end
```

or 

```elixir
# lib/foo.ex
defmodule Foo do
  def run() do
  end
end

defmodule Bar do
  def run() do
  end
end
```

lib/next_ls/db.ex
```diff
@@ -19,6 +19,9 @@ defmodule NextLS.DB do
   @spec insert_reference(pid(), map()) :: :ok
   def insert_reference(server, payload), do: GenServer.cast(server, {:insert_reference, payload})
 
+  @spec insert_reference(pid(), String.t()) :: :ok
+  def clean_references(server, filename), do: GenServer.cast(server, {:clean_references, filename})
```
mhanberg, line 23: spec is wrong

priv/monkey/_next_ls_private_compiler.ex
```diff
@@ -52,7 +52,13 @@ defmodule NextLSPrivate.Tracer do
 
   @source "user"
 
-  def trace(:start, _env) do
+  def trace(:start, env) do
+    Process.send(
+      parent_pid(),
+      {{:tracer, :start}, URI.parse(env.file).path},
```
mhanberg, line 58: i'm fairly certain this is just a file path already, so you don't have to parse it as a URI and then convert it back to a path again

dvic:
> @dvic Cristian beat you to it 🤣

hah that's okay, did not start working on it

@crbelaus awesome job :)!

crbelaus:
> One thing that I am unsure about with the `:start` trace is whether it is called only for a file, or if it is called multiple times if you define submodules or multiple modules in the same file
> 
> so, it should still work even if you have this code
> 
> ```elixir
> # lib/foo.ex
> defmodule Foo do
>   defmodule Bar do
>     def run() do
>     end
>   end
> 
>   def run() do
>   end
> end
> ```
> 
> or
> 
> ```elixir
> # lib/foo.ex
> defmodule Foo do
>   def run() do
>   end
> end
> 
> defmodule Bar do
>   def run() do
>   end
> end
> ```

My understanding after reading the documentation is that in those scenarios the `:start` callback will be called only once, but I will verify it later today and report back.

On the other hand I have a question about this @mhanberg: should we clean up the `symbols` table as well as the `references` table?

mhanberg:
I think the symbol purging is fine where it is, but I did put it where it is because I was un clear about how many times start is called.

mhanberg:
> but I will verify it later today and report back.

Thanks 🙇🏻

crbelaus:
@mhanberg I've just checked an even when having two modules on the same file and having nested modules the start is only called once per file when any of its contents change.

priv/monkey/_next_ls_private_compiler.ex
crbelaus: You are right. I've just removed the unused `URI.parse` call

lib/next_ls/db.ex
crbelaus: Ouch! Bad copy&paste 🤦‍♂️

mhanberg:
Excellent

lib/next_ls/db.ex
mhanberg, line 147:
```suggestion
```

PR #192, opened 2023-08-20
feat: auto update

mhanberg:
feat: auto update

Auto update is opt-in via an environment variable.

This env var will only be set by the editor extensions if it's
being started from the default location, so that if you
are installing NextLS via a package manager, it won't try to update
itself.


mhanberg:
* main
  * **PR #192** <a href="https://app.graphite.dev/github/pr/elixir-tools/next-ls/192" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/elixir-tools/next-ls/192?utm_source=stack-comment).

PR #199, opened 2023-08-22
simple fuzzy search of workspace symbols

biletskyy:
This is simple fuzzy search of Workspace symbols. It only check that all letters of query present in symbol name in exact order. There is no sorting right now. 

It uses case sensitive search if query contains a uppercase letter. If not then case insensitive search is used.

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -235,14 +235,7 @@ defmodule NextLS do
   end
 
   def handle_request(%WorkspaceSymbol{params: %{query: query}}, lsp) do
-    filter = fn sym ->
-      if query == "" do
-        true
-      else
-        # TODO: sqlite has a regexp feature, this can be done in sql most likely
-        to_string(sym) =~ query
-      end
-    end
+    case_sensitive? = String.downcase(query) != query
 
     symbols = fn pid ->
       rows =
@@ -270,7 +263,7 @@ defmodule NextLS do
 
     symbols =
       dispatch(lsp.assigns.registry, :databases, fn entries ->
-        for {pid, _} <- entries, symbol <- symbols.(pid), filter.(symbol.name) do
+        for {pid, _} <- entries, symbol <- symbols.(pid), fuzzy_match?(symbol.name, query, case_sensitive?) do
           name =
             if symbol.type != "defstruct" do
               "#{symbol.type} #{symbol.name}"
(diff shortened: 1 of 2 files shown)


mhanberg:
Can you share a screen recording of this in action?

biletskyy:
@mhanberg is that fine? I don't know how to make good recordings.

https://github.com/elixir-tools/next-ls/assets/190730/aea4e8b9-6a0f-4339-bef6-06eba926ea46

mhanberg:
This seems great and a strict improvement over the status quo!

I did notice one thing tho, it seems you have to include spaces for it to correctly match `def handle_info` with the query `def info`, so `definfo` doesn't work.

I am interested in merging this if you can smooth out that edge case.


https://github.com/elixir-tools/next-ls/assets/5523984/8e1fd0f0-eec7-427a-b227-c721992cf8f3

biletskyy:
I was puzzled by your findings. And I see several unrelated issues there. First of all the search only uses symbol name. So when you search for "def" it actually finds letters "def" in a "hanDlE_inFo" for example. But we send results back like `"#{symbol.type} #{symbol.name}"` and then VSCode tries to highlight matching letters and highlights first "DEF handle_info". 

The second issue is that if we use space in VSCode search then we don't receive anything after that space. I'm not sure where it gets lost. I'm still researching it. So if we repeat any search term with space it would work. But then VSCode filters visible results by its own. It's very confusing behavior with unpredicted results.

biletskyy:
as you can see VSCode continue to search same letters in search results. It also uses filename as search data.

https://github.com/elixir-tools/next-ls/assets/190730/c9df388f-6055-4033-b253-ee02001386e5

biletskyy:
I found the issue. VSCode breaks query to multi queries using whitespace. So we receive only first part of that and the rest of matching and filtering is done by vscode. I'm not sure how we should deal with it, i'm not that familiar with vscode extension yet, so not sure how hard it would be to fix that. But it looks like it is a work for vscode extension. 

image


https://github.com/microsoft/vscode/blob/3fac7f9a6be31b1c9895fa3f52c269d1900dbdad/src/vs/base/common/fuzzyScorer.ts#L850C45-L850C45

mhanberg:
So does it actually send multiple queries to the LSP or does it just use the next set of tokens to filter locally?

mhanberg:
Also, i'll check how it works in nvim later tonight hopefully

biletskyy:
> So does it actually send multiple queries to the LSP or does it just use the next set of tokens to filter locally?

It sends just first part before whitespace and uses other parts to locally filter results. For this local filtering it also uses meta data such as path which is shown in search results in VSCode.

I was looking for any documentation of this behaviour but was unable to find any. My guess is that it is done to be able to filter also by file path. And that makes sense since they assumed that symbols can't have whitespace in their names so it will be safe to use whitespace in different ways.

PR #208, opened 2023-08-28
fix: redirect logger to stderr

fhunleth:
This makes error logs show up in VSCode's error window without
interfering with communication over stdout.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,4 +2,6 @@ import Config
 
 config :next_ls, :indexing_timeout, 100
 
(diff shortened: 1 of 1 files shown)


config/config.exs
```diff
@@ -2,4 +2,6 @@ import Config
 
 config :next_ls, :indexing_timeout, 100
 
+config :logger, :default_handler, config: [type: :standard_error]
```
mhanberg, line 5: can you point me to the docs on this config option? I can't seem to find it in the OTP docs

PR #212, opened 2023-08-30
ranked fuzzy match search of workspace symbols

biletskyy:
This is an implementation of a sequential fuzzy string matching algorithm, similar to those used in code editors like Sublime Text. It is based on Forrest Smith's work on https://github.com/forrestthewoods/lib_fts/ and his blog post https://www.forrestthewoods.com/blog/reverse_engineering_sublime_texts_fuzzy_match/

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -235,14 +235,7 @@ defmodule NextLS do
   end
 
   def handle_request(%WorkspaceSymbol{params: %{query: query}}, lsp) do
-    filter = fn sym ->
-      if query == "" do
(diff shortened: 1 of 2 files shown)


biletskyy:
> Does this supercede your other PR?

yes. Sorry, forget to mention that.

mhanberg:
This is awesome, thanks!

A follow up could be to include the `def`, `defmodule`, etc when performing the match. 

That seemed to be a confusing point originally, and helps when searching for a like a macro vs a function, etc.

PR #152, opened 2023-08-09
feat: go to definition when aliases modules

mhanberg:
Collects references to modules in alias calls, which allows you do use go to definition and find references on them

## Example

```elixir
alias Foo.Bar.Baz
#     ^^^^^^^^^^^

alias Foo.Bar.{Baz, Bing}
#              ^^^  ^^^^

alias Foo.Bar.{
  Baz,
# ^^^
  Bing
# ^^^^
}
```

PR #215, opened 2023-09-13
feat: module attributes

### LINKED ISSUE #86: [definition] module attributes

biletskyy:
resolves #86 

Parses AST to collect module attribute definitions. Also parses AST to get attribute reference name.

### DIFF
--- a/lib/next_ls/runtime/sidecar.ex
+++ b/lib/next_ls/runtime/sidecar.ex
@@ -15,7 +15,7 @@ defmodule NextLS.Runtime.Sidecar do
   end
 
   def handle_info({:tracer, payload}, state) do
-    "Elixir." <> module_name = to_string(payload.module)
+    module_name = payload.module |> to_string() |> String.replace("Elixir.", "")
     all_symbols = parse_symbols(payload.file, module_name)
     attributes = filter_attributes(all_symbols)
 
--- a/lib/next_ls/runtime/sidecar.ex
+++ b/lib/next_ls/runtime/sidecar.ex
@@ -15,7 +15,30 @@ defmodule NextLS.Runtime.Sidecar do
   end
 
   def handle_info({:tracer, payload}, state) do
+    "Elixir." <> module_name = to_string(payload.module)
+    all_symbols = parse_symbols(payload.file, module_name)
+    attributes = filter_attributes(all_symbols)
+
+    payload = Map.put_new(payload, :symbols, attributes)
     DB.insert_symbol(state.db, payload)
+
+    {:noreply, state}
+  end
+
+  def handle_info({{:tracer, :reference, :attribute}, payload}, state) do
+    ast = payload.file |> File.read!() |> Code.string_to_quoted!(columns: true)
+    location = [line: payload.meta[:line], column: payload.meta[:column]]
+
+    {_ast, name} =
+      Macro.prewalk(ast, nil, fn
+        {:@, ^location, [{name, _meta, nil}]} = ast, _acc -> {ast, name}
+        other, acc -> {other, acc}
+      end)
+
+    if name do
+      payload = %{payload | identifier: "@#{name}"}
+      DB.insert_reference(state.db, payload)
+    end
 
     {:noreply, state}
   end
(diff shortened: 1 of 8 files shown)


biletskyy:
tested in VSCode

https://github.com/elixir-tools/next-ls/assets/190730/a847648b-a480-4d1c-b0ee-66d5db7341c4

lib/next_ls/runtime/sidecar.ex
```diff
@@ -15,11 +15,34 @@ defmodule NextLS.Runtime.Sidecar do
   end
 
   def handle_info({:tracer, payload}, state) do
+    module_name = payload.module |> to_string() |> String.replace("Elixir.", "")
```
mhanberg, line 18: You can use `Macro.to_string/1` instead of chopping off the `Elixir.` prefix

lib/next_ls/runtime/sidecar.ex
```diff
@@ -31,4 +54,60 @@ defmodule NextLS.Runtime.Sidecar do
 
     {:noreply, state}
   end
+
+  defp filter_attributes(symbols) do
+    symbols
+    |> Enum.filter(&match?({:attribute, _, _, _}, &1))
+    |> Enum.reject(fn {_, "@" <> name, _, _} ->
+      Map.has_key?(Module.reserved_attributes(), String.to_atom(name))
+    end)
```
mhanberg, line 63: For optimum efficiency, i'd combine these so its operated in a single pass

lib/next_ls/runtime/sidecar.ex
```diff
@@ -31,4 +54,60 @@ defmodule NextLS.Runtime.Sidecar do
...
+  defp parse_symbols(file, module) do
+    ast = file |> File.read!() |> Code.string_to_quoted!(columns: true)
+
+    {_ast, %{symbols: symbols}} =
+      Macro.traverse(ast, %{modules: [], symbols: []}, &prewalk/2, &postwalk(&1, &2, module))
+
+    symbols
+  end
+
+  # add module name to modules stack on enter
```
mhanberg, line 75: not gonna lie, i was not really sure as to the difference bewteen pre and post walk until i read this PR 😅

lib/next_ls/runtime/sidecar.ex
```diff
@@ -31,4 +54,60 @@ defmodule NextLS.Runtime.Sidecar do
...
+  defp prewalk({:defmodule, _, [{:__aliases__, _, modules} | _]} = ast, acc) do
+    modules_string =
+      modules
+      |> Enum.map(&Atom.to_string/1)
+      |> Enum.intersperse(".")
+      |> List.to_string()
```
mhanberg, line 81: If I'm reading this correctly, I think you can use [Module.concat/1](https://hexdocs.pm/elixir/Module.html#concat/1) for this

mhanberg: It'll be a alias tho (atom) rather than string, but I'm not sure they have to be strings for what you're doing.

lib/next_ls/runtime/sidecar.ex
mhanberg, line 1: I think it would be helpful to extra a module for single use AST functions, to get them out of this sidecar process, as it's just meant to be a sidecar/proxy.

You can then unit test them if you'd like (I don't think it's necessary for this PR, but as we discover more complicated edge cases, would be useful)

mhanberg (changes requested):
This is really, really good! Thank you so much for the contribution.

I left some comments, but they should be easy to address. I'd actually like to get this merged in quickly because you have introduced a pattern that I think will help with #104 and #152.

lib/next_ls/runtime/sidecar.ex
biletskyy: yes, agree. Will fix that.

lib/next_ls/runtime/sidecar.ex
biletskyy: yeah, you are write. I don't need strings at all. Will rewrite and simplify it.

lib/next_ls/runtime/sidecar.ex
biletskyy: yes, that's good idea. I extracted it. I'm not happy about module design but I think it could be refactored later with more usecases. 
I'm not sure about naming so if you like call it differently I don't mind

lib/next_ls/runtime/sidecar.ex
biletskyy: simplified that to ditch strings here

mhanberg (approved):

PR #219, opened 2023-09-16
feat: remove need of epmd

jozuas:
I've setup `epmdless` for the output of `mix release`.

I am not sure whether it's a good idea to make `next-ls` epmdless for development - it would involve a bit more work and would make remote shell connections a bit more challenging, namely, require a bunch of extra arguments. I've done some research and there does not seem to be a straightforward way to setup Erlang emulator flags in Elixir outside of `mix release` without using `ERL_FLAGS` environment variable.


mhanberg:
I think that you'll need to pass some of those VM args to (like `epmd_module`) to the Port that starts the runtime, or else it will still use epmd

... but now I'm also wondering if that will work, since epmdless won't be installed in the runtime 🤔.

Let me know how it goes

Also, thank you so much for this contribution! I was just waffling whether to tackle this or to tackle something else, and you helped me make my choice 👍

jozuas:
Messed up my fork, and as we discussed this work is insufficient.

PR #235, opened 2023-09-21
fix: gracefully handle injected attributes

### LINKED ISSUE #234: Is there a minimum supported elixir version?
We have a two different elixir projects - one that is using 1.14 and OTP25, and another that's stuck on 1.13 with OTP25.

The 1.14 project is smaller/simpler (~5k loc for Elixir) and seems to work with next-ls wonderfully! 

The 1.13 project is significantly larger (~100k loc of Elixir) and has some weirdness with private git dependencies. It seems to crash or struggle a bit. Syntax highlighting works, but not getting any formatting or error reporting

I wanted to first check if the Elixir version would be a factor here as I continue to debug. Possibly something about my setup and private git deps could be the problem?

mhanberg:
fix: gracefully handle injected attributes

Fixes #234

### DIFF
--- a/lib/next_ls/runtime/sidecar.ex
+++ b/lib/next_ls/runtime/sidecar.ex
@@ -60,8 +60,12 @@ defmodule NextLS.Runtime.Sidecar do
   end
 
   def handle_info({{:tracer, :reference, :attribute}, payload}, state) do
-    name = Attributes.get_attribute_reference_name(payload.file, payload.meta[:line], payload.meta[:column])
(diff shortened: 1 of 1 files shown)


mhanberg:
* main
  * **PR #235** <a href="https://app.graphite.dev/github/pr/elixir-tools/next-ls/235" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/elixir-tools/next-ls/235?utm_source=stack-comment).

Issue #245, opened 2023-09-28
Maybe support specifying MIX_TARGET and MIX_ENV


jjcarstens:
This may be outside the goals, but when working with Nerves projects it is sometimes nice to have the LS use a specific target so I can still see lib symbols that might be only used for that target.

I have initial POC in a fork, but I think it still needs some work to support specific targets (it currently takes reloading a couple times to get things to compile)

* https://github.com/jjcarstens/next-ls/tree/support-target-env
* https://github.com/jjcarstens/elixir-tools.vscode/tree/support-target-env

Adding in `MIX_ENV` was kind of a freebie and could probably be broken out and supported first (though not sure what demand for it is)

mhanberg:
Would you mind opening a draft PR for the next ls changes you made just so it's easier for me to see?

mhanberg:
Closed by #246

PR #246, opened 2023-09-29
Support passing `target` and `env` options from LSP

jjcarstens:
Based on #245

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -91,7 +91,12 @@ defmodule NextLS do
   @impl true
   def handle_request(
         %Initialize{
-          params: %InitializeParams{root_uri: root_uri, workspace_folders: workspace_folders, capabilities: caps}
+          params: %InitializeParams{
+            root_uri: root_uri,
+            workspace_folders: workspace_folders,
+            capabilities: caps,
+            initialization_options: init_opts
+          }
         },
         lsp
       ) do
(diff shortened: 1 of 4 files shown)


mhanberg:
This looks good, I think the second part would be adding `workspace/didChangeConfiguration` method so you can change the env and target at runtime without restarting the editor.

that can be done later tho.

I'll make some updates to this this weekend if you're cool with it and then i'll update the editor plugins.

thanks for the patch!

lib/next_ls.ex
```diff
@@ -91,7 +91,12 @@ defmodule NextLS do
   @impl true
   def handle_request(
         %Initialize{
-          params: %InitializeParams{root_uri: root_uri, workspace_folders: workspace_folders, capabilities: caps}
+          params: %InitializeParams{
+            root_uri: root_uri,
+            workspace_folders: workspace_folders,
+            capabilities: caps,
+            initialization_options: opts
```
mhanberg, line 98:
```suggestion
            initialization_options: init_opts
```

lib/next_ls.ex
```diff
@@ -511,6 +516,9 @@ defmodule NextLS do
 
     GenLSP.log(lsp, "[NextLS] Booting runtimes...")
 
+    target = lsp.assigns.opts["target"] || "host"
+    env = lsp.assigns.opts["env"] || "dev"
```
mhanberg, line 520: lets name these "mix_target" and "mix_env"

lib/next_ls.ex
```diff
@@ -124,7 +129,7 @@ defmodule NextLS do
          }
        },
        server_info: %{name: "Next LS"}
-     }, assign(lsp, root_uri: root_uri, workspace_folders: workspace_folders, client_capabilities: caps)}
+     }, assign(lsp, root_uri: root_uri, workspace_folders: workspace_folders, client_capabilities: caps, opts: opts)}
```
mhanberg, line 132: before pushing it into the assigns, I think we should process them via a safe list and do the defaulting.

PR #266, opened 2023-10-04
fix: don't fail when document is missing

mhanberg:
fix: don't fail when document is missing

There is a report of the document being nil, which is causing the server
to crash. The document should never be nil, but we can figure that out
later. The server not crashing should make that easier.

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -451,41 +451,52 @@ defmodule NextLS do
     document = lsp.assigns.documents[uri]
 
     [resp] =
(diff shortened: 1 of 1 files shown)


mhanberg:
* main
  * **PR #266** <a href="https://app.graphite.dev/github/pr/elixir-tools/next-ls/266" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/elixir-tools/next-ls/266?utm_source=stack-comment).

PR #268, opened 2023-10-04
feat: allow port from env

Cthuflu:
Provides an alternate way to set the tcp port (or a default) from an environment variable rather than strictly by argv.

### DIFF
--- a/lib/next_ls/lsp_supervisor.ex
+++ b/lib/next_ls/lsp_supervisor.ex
@@ -22,6 +22,13 @@ defmodule NextLS.LSPSupervisor do
       {opts, _, _invalid} =
         OptionParser.parse(argv, strict: [version: :boolean, help: :boolean, stdio: :boolean, port: :integer])
 
+      port =
+        opts[:port] || # fallback to env port if not set by args
+        case "NEXTLS_PORT" |> System.get_env("") |> Integer.parse() do
+          {int_port, ""} -> int_port
+          _bad_parse_or_error -> nil
+        end
+
       help_text = """
       Next LS v#{NextLS.version()}
 
@@ -60,9 +67,9 @@ defmodule NextLS.LSPSupervisor do
           opts[:stdio] ->
             []
 
-          is_integer(opts[:port]) ->
-            IO.puts("Starting on port #{opts[:port]}")
-            [communication: {GenLSP.Communication.TCP, [port: opts[:port]]}]
+          is_integer(port) ->
+            IO.puts("Starting on port #{port}")
+            [communication: {GenLSP.Communication.TCP, [port: port]}]
 
           true ->
             IO.puts(help_text)


mhanberg:
What's the use case?

Cthuflu:
To start next-ls with a port without needing to build command line arguments. My use-case is for starting from `iex -S mix` for development, however it can be useful in "prod" environments where argv may not be able to be passed.

It's more in-line with other configuration options existing as environment variables rather than command options (`NEXTLS_AUTO_UPDATE` and `NEXTLS_BINPATH`).

mhanberg:
I think i'd be more inclined just to include another bin script, eg `bin/iex` or `bin/start-iex`

```bash
#!/usr/bin/env bash

cd "$(dirname "$0")"/.. || exit 1

iex -S mix run --no-halt -e "Application.ensure_all_started(:next_ls)" -- "$@"
```

would that accomplish your use case?

Cthuflu:
Using scripts under bin is a less optimal way of working with mix and iex. I was going to also turn that script into a mix alias/task after this PR is processed.

Again; it would be useful to have an alternative (or a user-provided default) to dealing with argv.

mhanberg:
> Using scripts under bin is a less optimal way of working with mix and iex

can you elaborate?

I am mostly prodding at this because I am attempting to glean the utility of this patch, but I think you are assuming that your statement here is self-evident.

Cthuflu:
> can you elaborate?

Coming from Phoenix, rather than using a series of scripts under `bin/` for common tasks like starting the server; the project uses `mix` through custom tasks and aliases in development.

For example, some common tasks that are mix tasks or aliases:
- `mix phx.server`
  -  Starts the phoenix application and all servers in the tree
- `mix phx.digest` 
  - to digest and compress static files for release
- `mix ecto.setup`
  -  is an alias for running multiple mix tasks together, these are `ecto.create`, `ecto.migrate`, and `run priv/repo/seeds.exs`

Which can then be additionally ran through `iex`, such that `iex -S mix phx.server` lets the developer test with in REPL, manually `recompile()` without restarting the server if needed, or prod and observe the system as it is running.

These are tasks that are isolated in scope to working with Elixir, rather than explicitly writing scripts that will be platform dependent; even within linux not everyone has bash by default. Just include another script for single line commands like that only goes so far before you have a directory filled with them, and mix is _already_ working from the root directory that contains `mix.exs`. If you wanted to not litter scripts in that directory and try to unify their existing functionality you would then have to put together and deal with more complicated logic in _bash_, before even thinking of other platforms.

## How this interacts with the port argument; 

In development, starting the server is:

- `bin/start --port 9000`
- `bin/start_iex --port 9000`
  - Based on what you proposed
- `bin/start_iex_ash --port 9000`
  - Potential future, based on that
- `bin/start_iex_fish --port 9000`
  - Potential future, based on that
- `mix -- --port 9000`
- `iex -S mix -- --port 9000`

Now with the change in this PR, instead it is:

`export NEXTLS_PORT=9000` (or windows equivalent)

Now port is set as 9000 for all further commands.

- `bin/start` 
  - Start from the script (excluded from windows)
- `mix` 
  - Start with mix
- `iex -S mix` 
  - Start with mix in iex
- `mix nextls.server` or `iex -S mix nextls.server`
  -  no_halt and Application.ensure_all_started from `bin/start` can be stored here, in this theoretical task which is a later proposed change that is not in the scope of this discussion.

And this change works with the original script, such that the a default port can be passed.

`export NEXTLS_PORT=8999`

- `bin/start` 
  - Starts with tcp port 8999
- `bin/start --port 9000`  
  - Starts with port 9000 instead
- `bin/start --stdio` 
  - Starts as stdio instead
- `bin/start --port asdf` 
  - Will still error
- `NEXTLS_PORT=asdf bin/start` 
  - will still error
- `NEXTLS_PORT=asdf bin/start --port 9000` 
  - will start on 9000 since the command arg as been passed

It's more flexible and shouldn't get in the way of existing functionality.

mhanberg:
thanks for elaborating!

it seems to me like there isn't really a use case yet, as you explained the goal is more aligned with being more like phoenix (which i'm not sure entirely applies, as the phoenix mix tasks are used by a project consuming phoenix, not phoenix itself).

I'd be happy to reconsider if you find this extremely inconvenient during future contributions. but to me, it mostly seems like a way to not type a extra characters when starting the server locally.

Cthuflu:
It was something I found inconvenient during testing, with regards to how System.argv() was passing empty arrays instead of args when I directly started via mix and iex; as mix does have a bug with parsing arguments, which burrito ran into before.

Using `NEXTLS_PORT=9000 mix`  was meant to fix that issue.

I'll look into something else.

PR #277, opened 2023-10-05
Fix installation commands copying in the README.md

stefanluptak:
The `$` in front of the command was causing `command not found: $` error when using the copy button next to the snippet in the README.md file.


mhanberg:
Thanks!

PR #279, opened 2023-10-05
chore: improve patchelf block

jozuas:
I was a bit confused as I found that @mhanberg has already put patchelf code, but here's the work that I was unable to MR from last week.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -54,7 +54,7 @@
             erlang = beamPackages.erlang;
             elixir = beamPackages.elixir_1_15;
 
-            nativeBuildInputs = [ pkgs.xz pkgs.zig_0_11 pkgs._7zz ];
+            nativeBuildInputs = [ pkgs.xz pkgs.zig_0_10 pkgs._7zz pkgs.patchelf ];
 
             mixFodDeps = beamPackages.fetchMixDeps {
               inherit src version;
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -54,7 +54,7 @@
             erlang = beamPackages.erlang;
             elixir = beamPackages.elixir_1_15;
 
-            nativeBuildInputs = [ pkgs.xz pkgs.zig_0_11 pkgs._7zz ];
+            nativeBuildInputs = [ pkgs.xz pkgs.zig_0_10 pkgs._7zz pkgs.patchelf ];
```
mhanberg, line 57: We use zig 0.11 now, make sure to revert that part

flake.nix
```diff
@@ -83,16 +83,17 @@
...
+              cp -r ./burrito_out "$out"
 
+              if [ "${system}" == "x86_64-linux" ]; then
                 patchelf --set-interpreter ${pkgs.glibc}/lib/ld-linux-x86-64.so.2 "$out/burrito_out/next_ls_linux_amd64"
-              '' else ''
-                chmod +x ./burrito_out/*
-                cp -r ./burrito_out "$out"
-              '';
+              fi
+              if [ "${system}" == "aarch64-linux" ]; then
```
mhanberg, line 93: It didn't like when I built the flake on Darwin when you do the branching at the shell level, which is why I broke it out to the nix level

jozuas:
Really don't have time to work on Open Source right now after changing to a new, more demanding job.

mhanberg:
No worries, good luck!

Issue #45, opened 2023-06-26
completion


mhanberg:
## Features

Non exhaustive list of completion candiates

- [x] Modules
- [x] remote functions
- [x] local functions
- [x] private functions
- [x] variables/parameters
- [x] module attributes
- [x] structs
- [x] struct fields
- [x] bitstring modifiers
- [x] special forms
- [x] kernel functions
- [x] imported functions (eg, `import Foo`)
- [x] aliased modules (eg, `alias Foo.Bar` or `alias Foo.Bar, as: Baz`)

More features will be unlocked and made more robust as we work through https://github.com/elixir-lang/elixir/issues/12645

PR #289, opened 2023-10-17
feat: completions

mhanberg:
## Description

Implements basic autocompletion.

Current completion candidates supported

- Global modules
- Global Structs
- Struct fields
- Remote functions (w/ documentation)
- Special Forms
- Bitstring modifiers
- filesystem paths in strings

More features, particularly features that rely on contextual information about the code itself (meaning, which identifiers, aliases, imports are available) will come in subsequent patches.

Partially addresses #45 

## Experimental

This patch also introduces a new initialization option, `experimental`.

This feature will be gated as an experimental feature as it's built out. The purpose of this is so that early-early adopters can try it out and report bugs, but folks who would rather wait for something more stable won't have it affect their workflows.

To enable this feature, toggle the completions experiment in your editor.

### Nvim (elixir-tools.nvim)

```lua
require("elixir").setup({
    nextls = {
        enable = true,
        init_options = {
            experimental = {
                completions = {
                    enable = true
                }
            }
        }
    },
    elixirls = {enable = false}
})
```

### Visual Studio Code (elixir-tools.vscode)

```json
{
  "elixir-tools.nextLS.experimental.completions.enable": true
}
```

### Other editors

Not sure 😅

## Demos

TODO: record them my guy

## TODO

- [x] integration tests
- [ ] update elixir-tools.dev with instructions
- [x] update README with instructions

## Acknowedgements

This feature is initially based on `IEx.Autocomplete`. Huge thanks to the Elixir core team's efforts to help kickstart this feature. More deviations will likely occur as we gain more contextual parsing for things like imports, aliases and variables.

### DIFF
--- a/lib/next_ls/db.ex
+++ b/lib/next_ls/db.ex
@@ -195,12 +195,10 @@ defmodule NextLS.DB do
     case Exqlite.Basic.exec(conn, query, args) do
       {:error, %{message: message, statement: statement}, _} ->
         NextLS.Logger.warning(logger, """
-        ┌─────────────
-        │sqlite3 error: #{message}
-        │
-        │statement: #{statement}
-        │arguments: #{inspect(args)}
-        └─────────────
+        sqlite3 error: #{message}
+
+        statement: #{statement}
+        arguments: #{inspect(args)}
         """)
 
         {:error, message}
--- a/lib/next_ls/db.ex
+++ b/lib/next_ls/db.ex
@@ -195,10 +195,12 @@ defmodule NextLS.DB do
     case Exqlite.Basic.exec(conn, query, args) do
       {:error, %{message: message, statement: statement}, _} ->
         NextLS.Logger.warning(logger, """
-        sqlite3 error: #{message}
-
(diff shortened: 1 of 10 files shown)


lib/next_ls/db.ex
```diff
@@ -195,10 +195,12 @@ defmodule NextLS.DB do
...
-        sqlite3 error: #{message}
-
-        statement: #{statement}
-        arguments: #{inspect(args)}
+        ┌─────────────
+        │sqlite3 error: #{message}
+        │
+        │statement: #{statement}
+        │arguments: #{inspect(args)}
+        └─────────────
```
mhanberg, line 203: revert

lib/next_ls/helpers/ast_helpers/variables.ex
```diff
@@ -29,6 +29,18 @@ defmodule NextLS.ASTHelpers.Variables do
     end
   end
 
+  def collect(ast) do
```
mhanberg, line 32: comment out

PR #306, opened 2023-10-25
Cleanup same file occurrences when inserting a symbol

### LINKED ISSUE #305: Renaming a module leaves the old information in the database
Let's say we have the following module:

```elixir
defmodule MyModule do
  @moduledoc false
  def greet, do: "Hello"
end
```

The `symbols` table contains this:

```sql
sqlite> select * from symbols where module like 'MyModule%';
id     module    file                                                       type       name      line  column  source  inserted_at
-----  --------  ---------------------------------------------------------  ---------  --------  ----  ------  ------  -------------------
16064  MyModule  /Users/crbelaus/Developer/next-ls/lib/next_ls/mymodule.ex  defmodule  MyModule  1     1       user    2023-10-25 17:36:01
16065  MyModule  /Users/crbelaus/Developer/next-ls/lib/next_ls/mymodule.ex  def        greet     3     7       user    2023-10-25 17:36:01
```

Then I rename `MyModule` to `MyModuleRenamed`. Now the database contains this:

```sql
sqlite> select * from symbols where module like 'MyModule%';
id     module           file                                                       type       name             line  column  source  inserted_at
-----  ---------------  ---------------------------------------------------------  ---------  ---------------  ----  ------  ------  -------------------
16064  MyModule         /Users/crbelaus/Developer/next-ls/lib/next_ls/mymodule.ex  defmodule  MyModule         1     1       user    2023-10-25 17:36:01
16065  MyModule         /Users/crbelaus/Developer/next-ls/lib/next_ls/mymodule.ex  def        greet            3     7       user    2023-10-25 17:36:01
16066  MyModuleRenamed  /Users/crbelaus/Developer/next-ls/lib/next_ls/mymodule.ex  defmodule  MyModuleRenamed  1     1       user    2023-10-25 17:36:33
16067  MyModuleRenamed  /Users/crbelaus/Developer/next-ls/lib/next_ls/mymodule.ex  def        greet            3     7       user    2023-10-25 17:36:33
```

crbelaus:
This pull request ensures that when inserting a symbol we clean up the existing symbols for the same file.

Previously we were just removing the existing symbols for the same module, which worked if we update a module or rename its file. This left one corner case uncovered: renaming the module while keeping the filename.

This pull request fixes this and should fix #305. To test this you can perform the steps that I wrote in there and confirm that after renaming the module, the old module information is not present anymore in the `symbols` table.

### DIFF
--- a/lib/next_ls/db.ex
+++ b/lib/next_ls/db.ex
@@ -85,9 +85,9 @@ defmodule NextLS.DB do
       {conn, s.logger},
       ~Q"""
       DELETE FROM symbols
-      WHERE module = ?;
+      WHERE module = ? OR file = ?;
       """,
-      [mod]
+      [mod, file]
     )
 
     __query__(


crbelaus:
Marking as a draft as the tests are failing and I won't have time to fix it today.

crbelaus:
I am completely lost regarding the tests. I've cleaned everything and switched to `main` and they still fail on my machine. Not sure what I did 🙈

mhanberg:
@crbelaus i'll take a look tonight, no worries! Thanks for the hard work 💪

lib/next_ls/db.ex
```diff
@@ -85,9 +85,9 @@ defmodule NextLS.DB do
       {conn, s.logger},
       ~Q"""
       DELETE FROM symbols
-      WHERE module = ?;
+      WHERE module = ? OR file = ?;
```
mhanberg, line 88: I think this might need to be `AND`, because if two modules are written in the same file, either by nested `defmodule` or just two root `defmodule`s, the query written with the `OR` will clobber each other and delete symbols for another module.

But, I don't think that would solve the problem you are attempting to solve

mhanberg (commented):
I think the solution to this problem is going to be to add a new callback to the extension interface.

More or less, when we save a file, we should dispatch a "before_save" message to all extensions, so they can do any cleanup before we dispatch the compilation request to the runtimes

PR #314, opened 2023-10-29
Fix nextls binary name in installation guides

mickel8:
When downloading nextls binary from the `Releases` section on GH, it is named `next_ls`, while installation guides refer to `nextls`. In my case, I had to change vim lsp server registration function to use `next_ls` so maybe it's worth unifying the naming?


mhanberg:
Technically they are named `darwin_arm64_next_ls`, so I feel like the change here is still confusing.

but also `nextls` is how it is named when it is distributed everywhere else, so I wouldn't suggest changing all the docs anyway.

I would probably just add a comment about renaming it to `nextls` in the readme

PR #318, opened 2023-11-01
feat: to-pipe command

### LINKED ISSUE #55: to/from pipe
turn `Enum.to_list(map)` into `map |> Enum.to_list()` and vice versa.

NJichev:
Closes #55 

Aims to add to and from pipe to next-ls


lib/next_ls/commands/from_pipe.ex
```diff
@@ -0,0 +1,116 @@
+defmodule NextLS.Commands.FromPipe do
+  @moduledoc false
+  alias GenLSP.Enumerations.ErrorCodes
+  alias GenLSP.Structures.Position
+  alias GenLSP.Structures.Range
+  alias GenLSP.Structures.TextEdit
+  alias GenLSP.Structures.WorkspaceEdit
+
+  defp opts do
+    Schematic.map(%{
```
mhanberg, line 10: Please let me know if you have any feedback for the Schematic library! It can get the job done but it was first built with codegen in mind, so the ergonomics are a little off I think. Curious to hear your thought!

lib/next_ls/commands/from_pipe.ex
```diff
@@ -0,0 +1,116 @@
...
+      position: Schematic.map(%{character: Schematic.int(), line: Schematic.int()}),
```
mhanberg, line 11: For this, you can actually do `GenLSP.Structures.Position.schematic()` (i think)

test/next_ls/commands/from_pipe_test.exs
```diff
@@ -0,0 +1,160 @@
...
+               FromPipe.new(%{uri: uri, text: text, position: position})
+
+      assert edit.new_text == indent <> "Enum.to_list(map) |> Map.new()"
+      assert range.start.line == 2
+      assert range.start.character == 0
+      assert range.end.line == 2
+      assert range.end.character == String.length(expected_line)
+    end
+
+    test "works on separate lines when the coursor is on the pipe" do
```
mhanberg, line 70:
```suggestion
    test "works on separate lines when the cursor is on the pipe" do
```

test/next_ls/commands/from_pipe_test.exs
```diff
@@ -0,0 +1,160 @@
...
+      # When the coursor is on the pipe
```
mhanberg, line 71:
```suggestion
      # When the cursor is on the pipe
```

mhanberg: bunch of these, i'll let you get the rest

test/next_ls/commands/to_pipe_test.exs
```diff
@@ -0,0 +1,153 @@
...
+      assert range.end.line == line
+      assert range.end.character == String.length(expected_line)
+    end
+
+    @test_scenarios [
+      {"to_list(Map.new)", "Map.new() |> to_list()"},
+      {"to_list(a, b, c)", "a |> to_list(b, c)"},
+      {"Foo.Bar.baz(foo, bar)", "foo |> Foo.Bar.baz(bar)"},
+      {"Foo.Bar.baz(foo, bar, Map.new())", "foo |> Foo.Bar.baz(bar, Map.new())"}
+    ]
```
mhanberg, line 139: love this 🙌

test/next_ls/commands/to_pipe_test.exs
```diff
@@ -0,0 +1,153 @@
+defmodule NextLS.Commands.ToPipeTest do
+  use ExUnit.Case, async: true
+
+  alias GenLSP.Structures.TextEdit
+  alias GenLSP.Structures.WorkspaceEdit
+  alias NextLS.Commands.ToPipe
+
+  @moduletag :tmp_dir
+  @parse_error_code -32_700
```
NJichev, line 9: This reminds me to add test cases when I get bad args and when the code doesn't compile

lib/next_ls/commands/from_pipe.ex
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  defp from_pipe_edit(text, %{line: line}) do
+    with {:ok, lines, range_to_edit} <- find_pipe_lines(text, line),
+         {:ok, indent} <- get_indent(text, line),
+         {:ok, edit} <- get_edit(lines) do
+      {:ok, %TextEdit{new_text: indent <> edit, range: range_to_edit}}
+    else
+      {:error, _message} = error ->
+        error
```
mhanberg, line 75: I'd get rid of this, just let it blow up at the call site if any of the return values don't match

lib/next_ls/commands/from_pipe.ex
```diff
@@ -0,0 +1,116 @@
...
+  defp opts do
+    Schematic.map(%{
+      position: Schematic.map(%{character: Schematic.int(), line: Schematic.int()}),
+      uri: Schematic.str(),
+      text: Schematic.list(Schematic.str())
+    })
+  end
+
+  def new(opts) do
+    {:ok, %{text: text, uri: uri, position: position}} = Schematic.unify(opts(), Map.new(opts))
```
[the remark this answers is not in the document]
mhanberg: I was going to suggest wrapping the whole handle_event callback in the NextLS module in a try/rescue, so that we can log an error to the user and not crash the whole server.

You can publish a window/showMessage notification (I think there are examples in the codebase), so that it actually presents an UI notification to the user's editor

lib/next_ls.ex
```diff
@@ -592,6 +592,16 @@ defmodule NextLS do
       end
 
     {:reply, reply, lsp}
+  rescue
+    e ->
+      GenLSP.notify(lsp, %GenLSP.Notifications.WindowShowMessage{
+        params: %GenLSP.Structures.ShowMessageParams{
+          type: GenLSP.Enumerations.MessageType.warning(),
+          message: "[Next LS] #{command} command has crashed"
```
mhanberg, line 600: I would also log the exception, and the show message can be `[Next LS] #{command} has failed, see the logs for more details`

mhanberg:
Can you also write a couple of full integration tests?

mhanberg:
Okay, I tested this out and wasn't able to get it to work (after making some changes, I'll push them up on another branch so you can see, I rebased so i don't want to force push to your branch).

I think the integration tests will reveal where it breaks. I'll paste some of the errors I was seeing

to-pipe errors

```
NextLS: -32700: {[line: 1, column: 42], "missing terminator: } (for \"{\" starting at line 1)", ""}
```

from-pipe errors

```
NextLS: -32700: {[line: 1, column: 5], "syntax error before: ", "'|>'"}
```

mhanberg:
Changes i made: https://github.com/elixir-tools/next-ls/commit/fb33786927f8e986b460a2370848ca3b2fc40ea5

NJichev:
> Changes i made: [fb33786](https://github.com/elixir-tools/next-ls/commit/fb33786927f8e986b460a2370848ca3b2fc40ea5)

I have added the changes from this branch as well

test/support/utils.ex
```diff
@@ -114,6 +114,27 @@ defmodule NextLS.Support.Utils do
     end
   end
 
+  defmacro assert_opened(
```
NJichev, line 117: Not the best name, but I need to notify the server so that it loads the text for a certain URI into the documents assigns. Other things that I could do is to load the files preemptively with the initialization step. Another way would be to just expect the client to send the text as well? I think this is the cleanest though, just a little overhead in the tests.

mhanberg: seems fine, can you rename to `did_open` ?

and can you actually submit this as a separate PR, and add it to all the tests? then i can merge that and i won't have to approve your CI test runs every time you push

mhanberg:
i'll try and review/demo this before the weekend

mhanberg:
Tested this out, seems to break in quite a few cases.

I was able to turn `parser = code |> new(opts) |> next_token() |> next_token()` into `parser = next_token(next_token(new(code, opts)))`, but not the other way around. No error, just did nothing.

trying to do it on with the cursor on `parse_program` or `parser` here results in an error

```elixir
    case parse_program(parser) do
      {ast, %{errors: []}} ->
        {:ok, ast}

      {ast, %{errors: errors}} ->
        {:error, ast, errors}
    end
  end
```

The error being `NextLS: -32700: {[line: 1, column: 34], "missing terminator: end (for \"do\" starting at line 1)", ""}`

NJichev:
Thanks for testing it out, I'll add those as test scenarios.

The first scenario no idea why it didn't work, I'll have to take a look, also I'll do a better job of reporting the errors.
The second scenario I think it's an easy fix - I just have to add `do/end` as tokens to the bracket counting as well and I think it'll work fine, since currently it doesn't know to get more pieces of code until it compiles the `case`.
I will think of similar cases as this one

mhanberg:
You can use the error tolerant parser be been building (Spitfire) in Next LS now, so it might make this easier.

mhanberg:
https://github.com/elixir-tools/next-ls/assets/5523984/3e5d2da4-1e1b-414b-92a1-d079315ea5bf

found a case where it leaves a stray variable above the function it un-pipes

mhanberg:
Also that one was a little odd in that it un-piped the begging of the pipe chain instead of the one under the cursor.

I would have expected this more or less

```elixir
Enum.map(
  file |> NextLS.ASTHelpers.Variables.list_variable_references({line, col}),
  fn {_name, {startl..endl, startc..endc}} ->
    [file, startl, endl, startc, endc]
end)
```

I'm not sure if that is enough to stop this from merging, can improve it later

mhanberg:
also putting the cursor here results in the error

```elixir
#    👇
Enum.map(
  NextLS.ASTHelpers.Variables.list_variable_references(file, {line, col}),
  fn {_name, {startl..endl, startc..endc}} ->
    [file, startl, endl, startc, endc]
  end
)
```

`[Error] [NextLS] ** (ArithmeticError) bad argument in arithmetic expression`

mhanberg:
I'll see if i can't get some time to address those test cases tonight

mhanberg:
I have some local changes but haven't finished them yet, I'll try to get that done tomorrow.

mhanberg:
Fixing a test and the nix flake and then it'll be mergeable. 

I decided to include the Sourceror library as it has a lot of functions around positions that I was about to remake. As well as its zipper implementation which is useful

mhanberg:
Also I broke out from-pipe into a second PR

PR #319, opened 2023-11-01
ci: add cachix-action

akirak:
This PR is a follow up to #312 and is intended to:

- Save time on CI
- Make the binaries immediately available to Nix users

 by pushing newly built Nix assets to your cachix account.

As extra steps, you will have to:

- [x] Replace `YOUR_CACHIX_ACCOUNT` with the name of your cachix account.
- [x] Replace `YOUR_CACHIX_PUBLIC_KEY` which you can find at https://app.cachix.org/cache/YOUR_CACHIX_ACCOUNT
- [x] Generate a per-cache **write** auth token from https://app.cachix.org/cache/YOUR_CACHIX_ACCOUNT/settings/authtokens and add it as `CACHIX_AUTH_TOKEN` to the secrets of your repository (or even the organization).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -140,4 +140,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: nixbuild/nix-quick-install-action@v26
+      - uses: cachix/cachix-action@v12
+        with:
+          name: YOUR_CACHIX_ACCOUNT
+          signingKey: '${{ secrets.CACHIX_AUTH_TOKEN }}'
       - run: nix build
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -139,9 +139,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: nixbuild/nix-quick-install-action@v26
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@main
       - uses: cachix/cachix-action@v12
         with:
-          name: YOUR_CACHIX_ACCOUNT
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -128,4 +128,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: nixbuild/nix-quick-install-action@v26
+      - uses: cachix/cachix-action@v12
```
mhanberg, line 131: I think this should probably happen in the release workflow, so that it only caches new releases

mhanberg: so that nix users are only getting stable releases, not a canary/nightly build, and I think this would also cache binaries for pull requests? I guess i'm not entirely sure how the cache works

akirak: Cachix offers a decent amount of storage space for free (10 GiB at present), but if you want that level of control, it may be a good idea to use [magit-nix-cache-action](https://github.com/DeterminateSystems/magic-nix-cache-action/) instead for CI.

mhanberg: Can you elaborate on how this saves us CI time? we are using setup-beam for the elixir/erlang installation and already cache mix dependencies

akirak: It would save time on nix-build CI job I added in the previous PR. It's unnecessary if you don't intend to provide Nix packages from the flake. You are not using Nix to build the release assets, so the action would never accelerate your main workflow. Fro Nix, the action caches the Nix store, which saves time spent on rebuilding rebar3 with the latest Erlang every time.

PR #320, opened 2023-11-01
refactor: reduce imperative shell logic in flake.nix

shanesveller:
Regarding `HOME` specifically, I would've liked to eliminate that too as it's nearly always a pungent smell in Nix builds, but [Burrito.Steps.Build.PackAndBuild](https://github.com/burrito-elixir/burrito/blob/a60c6ab21156fc4c788907d33bfd0c546a022272/lib/steps/build/pack_and_build.ex#L26-L38) seems to
want a writable path to work in. AFAICT it's purely for scratch space, so we can just directly reuse the Nix build's temporary directory.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -58,16 +58,11 @@
             hash = "sha256-LV1DYmWi0Mcz1S5k77/jexXYqay7OpysCwOtUcafqGE=";
           };
 
-          preConfigure = ''
-            export HOME="$(pwd)"
-          '';
+          BURRITO_ERTS_PATH = "${beamPackages.erlang}/lib/erlang";
+          BURRITO_TARGET = lib.optional localBuild burritoExe.${system};
 
           preBuild = ''
-            export BURRITO_ERTS_PATH=${beamPackages.erlang}/lib/erlang
-          '';
-
-          preInstall = lib.optionalString localBuild ''
-            export BURRITO_TARGET="${burritoExe.${system}}"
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -29,7 +29,15 @@
       system,
       beamPackages,
       elixir,
-    }: rec {
+    }: let
+      aliased_7zz = pkgs.symlinkJoin {
+        name = "7zz-aliased";
+        paths = [pkgs._7zz];
+        postBuild = ''
+          ln -s ${pkgs._7zz}/bin/7zz $out/bin/7z
```
mhanberg, line 37: Oh heck yeah. I figured there had to be a nix way to do this

mhanberg (commented):
Looks nice, I'll take a closer look when I get back to a computer.

bin/7z
```diff
@@ -1,3 +0,0 @@
-#!/usr/bin/env bash
```
mhanberg, line 1: I still use in the release pipeline, so please revert

mhanberg:
Works on my x86 Linux and M1 Mac computers, will merge once you revert the bin/7z deletion

bin/7z
shanesveller: No problem at all! Reinstated and re-pushed at ff4c8175fe079d8502b82864fc3c6cd71bfa69d6.

PR #325, opened 2023-11-03
style(nix): remove use of Nix recursive definition syntax

akirak:
This PR gets rid of `rec` in flake.nix, following the advice by @shanesveller at https://github.com/elixir-tools/next-ls/pull/312#issuecomment-1789664824. It's only stylistic changes which won't change the behavior.


mhanberg:
thanks!

PR #326, opened 2023-11-03
ci: add dependabot.yml

akirak:
An older version of `actions/checkout` is used in one of your GitHub workflows. This PR sets up Dependabot which will notify you updates on used actions in the form of pull requests.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      timezone: UTC
+      interval: weekly
+      time: "11:00"
+    open-pull-requests-limit: 5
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      timezone: Asia/Kolkata
+      interval: weekly
+      time: "08:00"
```
mhanberg, line 8: is it possible to omit the timezone and time, or can you set it to UTC timezone with 11:00 for the time

akirak: Yes, you can set the interval/time to whatever you like.

akirak: Updated.

PR #324, opened 2023-11-02
ci: lint pull request title

mhanberg:

PR #327, opened 2023-11-03
ci: add workflow_dispatch

akirak:
My PRs (#325 and #326) failed in "Lint PR Title" step (added in #324) of the CI. I fixed those PR titles, but the statuses are not updated.

This PR adds a button like below, which you can use to manually dispatch the workflow without pushing a new commit to the branch.

wayshot-20231103-170933

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,7 @@ on:
   pull_request:
   push:
     branches: main
+  workflow_dispatch:
 
 jobs:
   commitlint:


mhanberg:
You can just push an empty commit to get them to re run. I squash and merge so an empty commit is not a big deal

akirak:
> just push an empty commit to get them to re run.

I personally wouldn't do that in my projects, but you can close this PR if you are fine with that way.

mhanberg:
I think there might be another pull_request type do list as a trigger that will run when the title changes actually. i need to read the docs on it.

mhanberg:
So, i think you want to add a new workflow to link the title, make the trigger "pull request" with types "open" and "edited"

akirak:
Please feel free to modify the commit or close this PR to open a new one.

PR #340, opened 2023-11-16
fix: sha for next-ls-deps

Eein:
Updates the sha for current next-ls-deps in the nix flake.


mhanberg:
apologies, i didn't get around to merging this, but i messed around with deps in another PR, so it supersedes this one

PR #348, opened 2023-12-20
refactor: diagnostic cache to an ets table

NJichev:
Refactors the Diagnostic Cache from using an Agent to a ETS table.

### DIFF
--- a/lib/next_ls/diagnostic_cache.ex
+++ b/lib/next_ls/diagnostic_cache.ex
@@ -1,33 +1,69 @@
 defmodule NextLS.DiagnosticCache do
-  # TODO: this should be an ETS table
   @moduledoc false
-  use Agent
+  use GenServer
 
   def start_link(opts) do
-    Agent.start_link(fn -> Map.new() end, Keyword.take(opts, [:name]))
+    opts = Keyword.take(opts, [:name])
+    GenServer.start_link(__MODULE__, opts, opts)
+  end
+
+  def init(opts) do
+    name = opts[:name]
+
+    ets =
+      if name do
+        :ets.new(name, [:named_table, :protected, :bag, {:read_concurrency, true}])
+      else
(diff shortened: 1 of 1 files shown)


lib/next_ls/diagnostic_cache.ex
```diff
@@ -1,33 +1,69 @@
...
+      end
+
+    {:ok, %{ets: ets}}
   end
 
   def get(cache) do
-    Agent.get(cache, & &1)
+    cache
+    |> GenServer.call(:all)
+    |> to_map()
```
NJichev, line 26: This is purely done to preserve the old interface of the cache, I originally refactored all the places that use it but decided to preserve it as is. Let me know what you think would be better, I think using the new list structure directly at some places might be a performance benefit other than that the old nested map structure is more convenient for use.

There was also the approach of storing it in ets as maps but it turned out to be a hassle since there's no update like interface like `Map.update`

mhanberg:
Thanks for taking a crack at this! I assume this is because of my TODO comment in the file.

Couple of things:

- Unless you read from the ets table directly, you aren't really going to see any read concurrency benefits, as its still bottle necked through the gen server.

- I added that comment originally naively, but i'm not sure if if this really is any kind of bottle neck.

There is now open telemetry in the project, complete with a docker-compose to spin up an otel collector and grafana tempo to view traces. It might be benficial to wrap the diagnostics cache in some spans and use it for a bit to see i the latency, then with an ets implementation and compare.

let me know if you want to try the tracing and i can help you through it.

NJichev:
I agree let's do it like this, I think I'll close this one for now then, since there are a couple of more improvements to be made, for example moving the reads to the table to the caller process, the problem I had was that tests spin up their own cache to keep the tests async so having them be separate processes made this harder. 

Maybe I could register the ets table in the lsp assigns as well to improve this so that I don't need to lookup the table ref every time.

Issue #334, opened 2023-11-08
Configure specific elixir and erlang paths/version for a runtime


mhanberg:
## Description

There is the use case that you might have a mono repo where each individual project is built on a different Elixir and OTP version. These should be accurately reflected by the runtime that is started for each sub project.

This will most likely be solved with the local config file #287.

PR #360, opened 2024-01-20
feat(runtime): support workspace shims with `elixir_exec` option

jjcarstens:
Relates to #334

IDEs like to work with shims rather than tools that alter path (such as `mise`/`rtx`). As a result, one must launch their IDE manually from a terminal session with all the `PATH` loaded in order to properly find the `elixir` executable. This works fine, but switching between projects/workspaces that have varying Elixir/OTP versions will most likely be using the version loaded when launching the program from terminal.

One resolve might be to include shims in the `PATH` globally. However, that negates some of the reasoning to move to alternate tools like `mise.

Instead, this allows setting the `elixir_exec` option to point directly at a shim. When used in workspace directories, that should also be able to load the correct versions defined in the tool version files. This also allows keeping the need for a shim isolated to the LS tool rather than global use

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -661,6 +661,7 @@ defmodule NextLS do
              uri: uri,
              mix_env: lsp.assigns.init_opts.mix_env,
              mix_target: lsp.assigns.init_opts.mix_target,
+             elixir_exec: lsp.assigns.init_opts.elixir_exec,
              on_initialized: fn status ->
                if status == :ready do
                  Progress.stop(lsp, token, "NextLS runtime for folder #{name} has initialized!")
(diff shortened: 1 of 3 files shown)


mhanberg:
I think this is semantics, but are you intending to set this to the path of the shim or the path of the actual binary?

jjcarstens:
It coil be either. Personally, I would set to shim because I use `mise` and that is static. The shim does evaluation of which elixir version to use and where it is located

mhanberg:
Does the shim read from the file or env vars. It's unclear to me if the shim would work still if the editor hasn't inherited the environment.

jjcarstens:
Shims are usual an executable or small script that's supposed to know how to find and forward to the target. So yes it should. 

In my little test I had initially cleared the environment and only gave the shim, but that affects the path globally and would break the other tests. Though running individually it seemed to work fine

PR #373, opened 2024-02-21
feat: autocomplete for kernel functions

lucacervello:
Add autocomplete for Kernel functions like `put_in/2`, `++/2`, `defstruct/1`


mhanberg:
So I originally didn't include Kernel functions because, unlike SpecialForms, you can't remove them from scope with `import`.

So here technically you could un-import `def` and make your own def and the autocomplete would show you the wrong documentation, or show it to you even tho you un-imported it (in the case you don't define your own.

That being said, it since completions are still experimental, it might be already to include Kernel functions, as for most people they probably aren't messing with them outside of library code.

mhanberg:
alright i played with this for 8 seconds and its worth adding

PR #375, opened 2024-02-22
feat: add require code action

NJichev:
This adds a require code action that adds `require Module` to your module whenever a macro is used without requiring it beforehand.
It tries to insert the require after all the top level Elixir macros(moduledoc, alias, require, import).

### DIFF
--- a/lib/next_ls/extensions/elixir_extension/code_action/require.ex
+++ b/lib/next_ls/extensions/elixir_extension/code_action/require.ex
@@ -10,7 +10,7 @@ defmodule NextLS.ElixirExtension.CodeAction.Require do
 
   @one_indentation_level "  "
   @spec new(diagnostic :: Diagnostic.t(), [text :: String.t()], uri :: String.t()) :: [CodeAction.t()]
-  def new(diagnostic = %Diagnostic{}, text, uri) do
+  def new(%Diagnostic{} = diagnostic, text, uri) do
     range = diagnostic.range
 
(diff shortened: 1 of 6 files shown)


test/next_ls/extensions/elixir_extension/code_action/unused_variable_test.exs
```diff
@@ -9,14 +9,18 @@ defmodule NextLS.ElixirExtension.UnusedVariableTest do
...
   test "adds an underscore to unused variables" do
-    text = """
-    defmodule Test.Unused do
-      def hello() do
-        foo = 3
-        :world
-      end
-    end
-    """
+    text =
```
NJichev, line 12: We get the text as a list of lines, which for the unused variable code isn't a problem but it's better to be consistent if anyone copies simple tests from other files like me.

lib/next_ls/extensions/elixir_extension/code_action/require.ex
```diff
@@ -0,0 +1,134 @@
...
+        {_, context, _} = Enum.max_by(top_level_macros, fn {_, ctx, _} -> ctx[:line] end)
+        context
+    end
+  end
+
+  defp get_module_name({:defmodule, _, [{:__aliases__, _, alias} | _rest]}) do
+    name =
+      alias
+      |> Module.concat()
+      |> to_string()
```
mhanberg, line 127: If you use Macro.to_string it doesn't include the Elixir. prefix

PR #376, opened 2024-02-24
test: fix failing tests

NJichev:
Fixes 2 cases of failing tests:
1) end range of one liner def was characters long as the start where I would assume it should be the characters that add up to the end, i.e `(1 level of indentation)def run, do: :ok` is 18 characters.
2) changed message format in the credo extension


mhanberg:
I fixed the Credo thing on main.

I'm not sure what the other thing with document symbols is. Those pass for me locally and they seem to have failed here on CI.

PR #385, opened 2024-02-28
feat: add snippets

### LINKED ISSUE #59: snippets

lucacervello:
Closes #59

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -602,8 +602,7 @@ defmodule NextLS do
 
         case NextLS.Snippet.get(label, nil) do
           nil -> [completion_item | results]
-          snippets when is_list(snippets) -> Enum.map(snippets, &Map.merge(completion_item, &1)) ++ results
-          snippet -> [Map.merge(completion_item, snippet) | results]
+          %{} = snippet -> [Map.merge(completion_item, snippet) | results]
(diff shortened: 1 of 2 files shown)


lib/next_ls.ex
```diff
@@ -592,12 +592,19 @@ defmodule NextLS do
...
+          %GenLSP.Structures.CompletionItem{
+            label: label,
+            kind: kind,
+            insert_text: name,
+            documentation: docs
+          }
+
+        case NextLS.Snippet.get(name, nil) do
+          nil -> [completion_item]
+          snippets when is_list(snippets) -> [completion_item] ++ snippets
```
mhanberg, line 605: I think the snippet should replace the completion item in most cases. 

Otherwise multiple items for let's say defmodule will be present.

mhanberg:
Another thing is you need to preserve the documentation in the completion item in the snippet

lucacervello:
Should I add a new parameter to the `Snippet.get/2` function or use the second one as docs?
Or we don't want to `Snippet.get/2` to change?

mhanberg:
I'll try and play with this soon, currently wrapping up some work on the editor extensions.

mhanberg:
made a couple changes, otherwise great!

PR #386, opened 2024-02-28
feat: add alias refactor workspace command

NJichev:
This adds an alias refactor workspace command to NextLS.

Your cursor should be at the target module that you wish to alias. It will insert the alias at the top of the nearest defmodule definition scoping the refactor only to the current module instead of the whole file.


test/next_ls/commands/alias_test.exs
```diff
@@ -0,0 +1,197 @@
...
+      expected_edit =
+        String.trim("""
+        defmodule MyApp do
+          alias Foo.Bar
+
+          def to_list(map) do
+            Bar.to_list(map)
+          end
+
+          def bar do
```
NJichev, line 76: Not sure why Macro.to_string picks to do/end block the one-liner. The ast from a one liner and a do/end block is the same so that's probably why. I might be missing some context data in one of the cases when testing in iex.

lib/next_ls.ex
```diff
@@ -645,6 +646,19 @@ defmodule NextLS do
...
           })
 
+        "alias-refactor" ->
+          [arguments] = params.arguments
+
+          uri = arguments["uri"]
+          position = arguments["position"]
+          text = lsp.assigns.documents[uri]
+
+          NextLS.Commands.Alias.refactor(%{
```
mhanberg, line 656: lets just call the "main method" function of these modules "run"

mhanberg:
@NJichev is this ready for review?

NJichev:
It's ready for review(s) now

mhanberg:
So I tried it out and was seeing some exceptions thrown.

To use AST with Sourceror, you need to pass the same literal_encoder as we're doing in the pipe command. This is because literals don't contain metadata (as they are literals), so we use the literal encoder to turn them into normal ast nodes that have the metadata.

I made that change and now the tests aren't passing, so you'll want to make that change and then get the tests passing again.

lib/foo.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Foo do
```
mhanberg, line 1: Was this file accidentally added?

NJichev: Yep my bad

mhanberg:
Ready for another review?

mhanberg:
@NJichev have you tested this in an editor? I can't seem to get it to work at all now.

mhanberg:
ahh, it doesn't work with a remote call with args, like `Foo.Bar.Baz.run()`

mhanberg:
You'll need to modify this to work with structs and also inside `import Foo.Bar.Baz` and `require Foo.Bar.Baz`

NJichev:
Those issues should be fixed now

mhanberg:
https://github.com/elixir-tools/next-ls/assets/5523984/e1243d8f-07b3-4cfb-9baa-bd4f358a3523

found another bug. tried to alias this struct and it added the alias to a submodule below and didn't actually change any of the references

mhanberg:
> found another bug. tried to alias this struct and it added the alias to a submodule below and didn't actually change any of the references

i noticed that this happens for the require module code action as well, if you get some time, please fix that in another PR as well.

NJichev:
I think I already fixed it here, the only work left here is to update spitfire and handle the comments metadata.

So the issue was that the find nearest module did literally this, and it found the nearest defmodule instead of the defmodule context of the cursor. Unless it's another issue.

 I'll take a look during the weekend and finish this PR.

mhanberg:
> I think I already fixed it here, the only work left here is to update spitfire and handle the comments metadata.
> 
> So the issue was that the find nearest module did literally this, and it found the nearest defmodule instead of the defmodule context of the cursor. Unless it's another issue.
> 
> I'll take a look during the weekend and finish this PR.

oh gotcha, cool. i'll check again once we get this merged

mhanberg:
Latest issue (after rebasing on main):

it seems to be putting the alias in the right module, but it seems to be duplicating some comments and shifting them into a weird spot


https://github.com/elixir-tools/next-ls/assets/5523984/d9ff03a4-d075-4846-bd65-78f87c76fb63

mhanberg:
Okay, my previous comment seems fixed and is working well, now a new problem 😅

it appears that new line escape sequences in strings are getting replaced with actual new lines

I need to test this in vscode as well to make sure this isn't a client issue

CleanShot 2024-04-18 at 08 24 46@2x

mhanberg:
Glorious!

PR #396, opened 2024-03-12
feat: signature help

lucacervello:
I've reduced the scope of this MR.

~I've reused the `Definition` code and showed the signature.~

It doesn't work while you're writing code like `String.replace(` ~and it doesn't show the active parameter.~

https://github.com/elixir-tools/next-ls/assets/5053083/1541b481-344d-45f1-b167-22ee983a4944

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [0.22.3](https://github.com/elixir-tools/next-ls/compare/v0.22.2...v0.22.3) (2024-05-15)
+
+
+### Bug Fixes
+
+* ensure some elixir internals are ready ([#478](https://github.com/elixir-tools/next-ls/issues/478)) ([f4685d0](https://github.com/elixir-tools/next-ls/commit/f4685d01266b4afb7f557d9a361fc7770aa22ec6)), closes [#467](https://github.com/elixir-tools/next-ls/issues/467)
+
+## [0.22.2](https://github.com/elixir-tools/next-ls/compare/v0.22.1...v0.22.2) (2024-05-14)
+
+
+### Bug Fixes
+
+* **completions:** work in guards ([#475](https://github.com/elixir-tools/next-ls/issues/475)) ([e0573ab](https://github.com/elixir-tools/next-ls/commit/e0573ab23c439313ed2546015f12a21dfe573d1d))
+
 ## [0.22.1](https://github.com/elixir-tools/next-ls/compare/v0.22.0...v0.22.1) (2024-05-13)
 
 
--- a/flake.nix
+++ b/flake.nix
@@ -14,7 +14,7 @@
   }: let
     inherit (nixpkgs) lib;
 
-    version = "0.22.1"; # x-release-please-version
+    version = "0.22.3"; # x-release-please-version
 
     # Helper to provide system-specific attributes
     forAllSystems = f:
--- a/lib/next_ls/autocomplete.ex
+++ b/lib/next_ls/autocomplete.ex
@@ -636,24 +636,6 @@ defmodule NextLS.Autocomplete do
       for {fun, arity} <- funs,
           name = Atom.to_string(fun),
           if(exact?, do: name == hint, else: String.starts_with?(name, hint)) do
-        # doc =
-        #   Enum.find(fdocs, fn {{type, fname, _a}, _, _, _doc, _} ->
-        #     type in [:function, :macro] and to_string(fname) == name
-        #   end)
-
-        # doc =
-        #   case doc do
-        #     {_, _, _, %{"en" => fdoc}, _} ->
-        #       """
-        #       ## #{Macro.to_string(mod)}.#{name}/#{arity}
-
-        #       #{NextLS.DocsHelpers.to_markdown(content_type, fdoc)}
-        #       """
-
-        #     _ ->
-        #       nil
-        #   end
-
         %{
           kind: :function,
           data: {mod, name, arity},
--- a/lib/next_ls/runtime.ex
+++ b/lib/next_ls/runtime.ex
@@ -231,22 +231,32 @@ defmodule NextLS.Runtime do
              true <- connect(node, port, 120) do
           NextLS.Logger.info(logger, "Connected to node #{node}")
 
-          :next_ls
-          |> :code.priv_dir()
-          |> Path.join("monkey/_next_ls_private_compiler.ex")
-          |> then(&:rpc.call(node, Code, :compile_file, [&1]))
-          |> tap(fn
-            {:badrpc, error} ->
-              NextLS.Logger.error(logger, "Bad RPC call to node #{node}: #{inspect(error)}")
-              send(me, {:cancel, error})
+          result =
+            :next_ls
+            |> :code.priv_dir()
+            |> Path.join("monkey/_next_ls_private_compiler.ex")
+            |> then(fn path ->
+              if await_config_table(node, 5) do
+                :rpc.call(node, Code, :compile_file, [path])
+              else
+                {:badrpc, "internal ets table not found"}
+              end
+            end)
+            |> then(fn
+              {:badrpc, error} ->
+                NextLS.Logger.error(logger, "Bad RPC call to node #{node}: #{inspect(error)}")
+                send(me, {:cancel, error})
+                :error
 
-            _ ->
-              :ok
-          end)
+              _ ->
+                :ok
+            end)
 
-          {:ok, _} = :rpc.call(node, :_next_ls_private_compiler, :start, [])
+          if result == :ok do
+            {:ok, _} = :rpc.call(node, :_next_ls_private_compiler, :start, [])
 
-          send(me, {:node, node})
+            send(me, {:node, node})
+          end
         else
           error ->
             send(me, {:cancel, error})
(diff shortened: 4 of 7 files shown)


lib/next_ls/signature_help.ex
```diff
@@ -0,0 +1,117 @@
...
+          code = File.read!(file)
+
+          params =
+            code
+            |> ASTHelpers.Functions.get_function_params(label, line, col)
+            |> Enum.map(fn {name, _, _} ->
+              %ParameterInformation{
+                label: Atom.to_string(name)
+              }
+            end)
```
mhanberg, line 33: I think an easier route here, to avoid having to parse the source code and find the function, is to embed the parameters into the database. I believe this information should be available in the compiler tracer.

mhanberg: That can be done is a separate PR as well as its an orthogonal chane

mhanberg: https://github.com/elixir-tools/next-ls/blob/main/lib/next_ls/db.ex#L85

mhanberg:
Also I had the thought, this needs to work for standard library and Erlang functions as well. Which those are traced by the compiler cuz they're already compiled. 

For those, I believe will need to read them from the runtime in some way (I'll leave you to figure that out)

I think we can ship a first version without that tho

mhanberg:
Played with it and realized that I'm not sure this path forward will work.

this searched for the definition of the module/function before the cursor, but that implies the project has compiled and its in the references table.

I think want we actually want to do is extract the module and function before the cursor and then fetch the docs with Code.fetch_docs/1

that should return the function docs which will actually have a "signature" field that we can use

CleanShot 2024-03-26 at 15 10 59@2x

lucacervello:
@mhanberg I switched approach.

It works well for something like

```elixir
Enum.map(list, fun)
``` 

It doesn't work for aliased or imported functions like

```elixir
alias NextLS.Formatter

Formatter.format(code)
```

Because `Code.fetch_docs` doesn't find `Formatter`

mhanberg:
Gotcha, that should become available as I build out the "buffer environment". 

I'll try this out today

lucacervello:
> Gotcha, that should become available as I build out the "buffer environment".
> 
> I'll try this out today

Thank you, if you're happy with that limitation, I can cleanup the code later 👍

mhanberg:
Yeah the limitation is fine for now, it's basically already a limitation for autocomplete.

lib/next_ls/signature_help.ex
```diff
@@ -0,0 +1,94 @@
...
+  alias GenLSP.Structures.ParameterInformation
+  alias GenLSP.Structures.SignatureHelp
+  alias GenLSP.Structures.SignatureInformation
+  alias Sourceror.Zipper
+
+  def fetch_mod_and_name(uri, position) do
+    with {:ok, text} <- File.read(URI.parse(uri).path),
+         ast =
+           text
+           |> Spitfire.parse()
```
mhanberg, line 15: You've got to set the literal encoder, you can see an example elsewhere in the codebase. I'm currently seeing errors, while trying to test this out and I'm pretty sure this is why.

lib/next_ls/signature_help.ex
```diff
@@ -0,0 +1,94 @@
...
+      end
+    end
+  end
+
+  def format({:ok, {:docs_v1, _, :elixir, _, _, _, docs}}, func_name) do
+    docs
+    |> Enum.filter(fn
+      {{_, name, _arity}, _, _, _, _} -> name == func_name
+    end)
+    |> Enum.map(fn
```
mhanberg, line 32: for speed, lets combine the filter and map operations, can easily be done with a comprehensions

```elixir
for {{_, name, _arity}, _, [signature], _, _} <- docs, name == func_name do
  # the code
end
```

lib/next_ls/signature_help.ex
```diff
@@ -0,0 +1,94 @@
...
+          signature
+          |> Spitfire.parse!()
+          |> then(fn {_, _, args} ->
+            Enum.map(args, fn {name, _, _} -> name end)
+          end)
+          |> Enum.map(fn name ->
+            %ParameterInformation{
+              label: Atom.to_string(name)
+            }
+          end)
```
mhanberg, line 44: another speed one, the two maps can be combined
```suggestion
          |> then(fn {_, _, args} ->
            Enum.map(args, fn {name, _, _} -> 
              %ParameterInformation{
                label: Atom.to_string(name)
              }
            end)
          end)
```

lib/next_ls/signature_help.ex
```diff
@@ -0,0 +1,94 @@
...
+              documentation: %MarkupContent{
+                kind: MarkupKind.markdown(),
+                value: ""
+              }
+            }
+          ]
+        }
+
+      # {{_, _name, _arity}, _, [], _, _} ->
+      #   []
```
mhanberg, line 60: commented out code

lib/next_ls/signature_help.ex
```diff
@@ -0,0 +1,94 @@
...
+        _ ->
+          false
+      end)
+
+    if result do
+      {:ok, Zipper.node(result)}
+    else
+      {:error, :not_found}
+    end
+  end
```
mhanberg, line 93: this might be a candidate to move to the ast helpers module, something like "find_within"

lib/next_ls/signature_help.ex
```diff
@@ -0,0 +1,94 @@
...
+          end)
+
+        %SignatureHelp{
+          signatures: [
+            %SignatureInformation{
+              label: signature,
+              parameters: params_info,
+              documentation: %MarkupContent{
+                kind: MarkupKind.markdown(),
+                value: ""
```
mhanberg, line 53: we want the documentation in there

lib/next_ls/signature_help.ex
lucacervello: I went with `find_aliased_function_call_within`, because I wanted to be very clear that the function has limitations. Happy to change the name to a more meaningful one if you have any

mhanberg: ahh, yes i see that now.

"remote function" is the "more correct" term than "aliased"

lib/next_ls/signature_help.ex
```diff
@@ -0,0 +1,64 @@
...
+
+  alias GenLSP.Enumerations.MarkupKind
+  alias GenLSP.Structures.MarkupContent
+  alias GenLSP.Structures.ParameterInformation
+  alias GenLSP.Structures.SignatureHelp
+  alias GenLSP.Structures.SignatureInformation
+  alias NextLS.ASTHelpers
+
+  def fetch_mod_and_name(uri, position) do
+    with {:ok, text} <- File.read(URI.parse(uri).path),
```
mhanberg, line 12: you'll want to pass in the document from the handle_request callback (should be something like `lsp.assigns.documents[uri] |> Enum.join("\n")`.

and in the tests, use a didOpen and then didChange notifcation to get the server state ready to send the signature help request.

mhanberg:
Also, I think we should go ahead and make the active parameter work before merging. 

I imagine you can take the ast you find and then find the parameter the cursor is inside and use that to determine the index to use.

lucacervello:
@mhanberg I added the active parameter, tests are working but I'm having trouble using it in my editor. I'll investigate more later.

mhanberg:
CleanShot 2024-04-12 at 13 19 19@2x
can confirm that the active parameter seems to not be working in the editor, but i rebased on main (i just merged a big change #410 ) and it works other than that, for stdlib, deps, and project functions

mhanberg:
I think i need to make a patch to spitfire for this to work the way its currently coded.

mhanberg:
I was re reading the PR comments and I can't seem to remember what patch I needed to make to spitfire.

lucacervello:
> Now spitfire has the function `Spitfire.container_cursor_to_quoted`, which is used to insert a `__cursor__()` marker in the ast where the cursor is.

This could work but only for partial code like

```elixir
Enum.map(
```

I think we also want to give `SignatureHelp` when the cursor is on valid code like 

```elixir
Enum.map([1, 2], fn n -> n + 1 end)
```

mhanberg:
> > Now spitfire has the function `Spitfire.container_cursor_to_quoted`, which is used to insert a `__cursor__()` marker in the ast where the cursor is.
> 
> This could work but only for partial code like
> 
> ```elixir
> Enum.map(
> ```
> 
> I think we also want to give `SignatureHelp` when the cursor is on valid code like
> 
> ```elixir
> Enum.map([1, 2], fn n -> n + 1 end)
> ```

You make it into a partial code snippet by truncating the document after the cursor position.

Basically if the cursor is the location line: 0, character: 30, you'd truncate it too look like `Enum.map([1, 2], fn n -> n + ` and then run the following operations, which produces ast you can use

```
iex(2)> {:ok, ast} = Spitfire.container_cursor_to_quoted("Enum.map([1, 2], fn n -> n + ")
{:ok,
 {{:., [line: 1, column: 5],
   [
     {:__aliases__, [last: [line: 1, column: 1], line: 1, column: 1], [:Enum]},
     :map
   ]}, [closing: [line: 1, column: 45], line: 1, column: 6],
  [
    [1, 2],
    {:fn, [closing: [line: 1, column: 42], line: 1, column: 18],
     [
       {:->, [line: 1, column: 23],
        [
          [{:n, [line: 1, column: 21], nil}],
          {:+, [line: 1, column: 28],
           [
             {:n, [line: 1, column: 26], nil},
             {:__cursor__,
              [closing: [line: 1, column: 41], line: 1, column: 30], []}
           ]}
        ]}
     ]}
  ]}}
iex(3)> ast |> Macro.to_string() |> IO.puts
Enum.map([1, 2], fn n -> n + __cursor__() end)
```

lucacervello:
> You make it into a partial code snippet by truncating the document after the cursor position.

Oh right, It make a lot of sense, dumb question, my bad 😅. Thank you.

I hope I have some time in the upcoming weeks to push this PR forward 🤞

mhanberg:
> > You make it into a partial code snippet by truncating the document after the cursor position.
> 
> Oh right, It make a lot of sense, dumb question, my bad 😅. Thank you.
> 
> I hope I have some time in the upcoming weeks to push this PR forward 🤞

No worries, its all volunteer work! Work on it when you have the time, energy, and interest.

PR #397, opened 2024-03-13
feat: add params to symbols table 

lucacervello:


mhanberg:
You can store them as a blob using termtobinary and then after querying them do binary to term

mhanberg:
Fix the tests and then I'll review

lib/next_ls/db.ex
```diff
@@ -127,14 +127,23 @@ defmodule NextLS.DB do
       )
     end
 
-    for {name, {:v1, type, _meta, clauses}} <- defs, {meta, _, _, _} <- clauses do
+    for {name, {:v1, type, _meta, clauses}} <- defs, {meta, args, _, _} <- clauses do
```
mhanberg, line 130: Rename args to params

PR #398, opened 2024-03-14
feat: defmodule snippet infer module name

lucacervello:
defmodule snippet infer module name

recording 🎥 

https://github.com/elixir-tools/next-ls/assets/5053083/e5e6aaa4-5a63-467b-b41f-49f6fd35a47a

### DIFF
--- a/lib/next_ls/helpers/ast_helpers/env.ex
+++ b/lib/next_ls/helpers/ast_helpers/env.ex
@@ -113,6 +113,8 @@ defmodule NextLS.ASTHelpers.Env do
     }
   end
 
+  def ascend(nil, acc, _callback), do: acc
+
   def ascend(%Zipper{path: nil} = zipper, acc, callback), do: callback.(Zipper.node(zipper), zipper, acc)
 
   def ascend(zipper, acc, callback) do
--- a/lib/next_ls/snippet.ex
+++ b/lib/next_ls/snippet.ex
@@ -1,19 +1,30 @@
 defmodule NextLS.Snippet do
   @moduledoc false
 
-  def get("defmodule/2", nil) do
+  def get(label, trigger_character, opts \\ [])
+
+  def get("defmodule/2", nil, opts) do
+    uri = Keyword.get(opts, :uri)
+
+    modulename =
+      if uri do
+        infer_module_name(uri)
+      else
+        "ModuleName"
+      end
+
     %{
       kind: GenLSP.Enumerations.CompletionItemKind.snippet(),
       insert_text_format: GenLSP.Enumerations.InsertTextFormat.snippet(),
       insert_text: """
-      defmodule ${1:ModuleName} do
+      defmodule ${1:#{modulename}} do
         $0
       end
       """
     }
   end
 
-  def get("defstruct/1", nil) do
+  def get("defstruct/1", nil, _opts) do
     %{
       kind: GenLSP.Enumerations.CompletionItemKind.snippet(),
       insert_text_format: GenLSP.Enumerations.InsertTextFormat.snippet(),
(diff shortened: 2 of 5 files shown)


lib/next_ls/snippet.ex
```diff
@@ -171,7 +182,47 @@ defmodule NextLS.Snippet do
...
-  def get(_label, _trigger_character) do
+  def get(_label, _trigger_character, _opts) do
     nil
   end
+
+  defp infer_module_name(uri) do
+    result =
+      uri
+      |> Path.split()
+      |> Enum.reduce(false, fn
```
mhanberg, line 193: Can you explain what the first element of the accumulator signifies here?

The clause with support is confusing me a tad.

mhanberg:
You're on fire!

lib/next_ls/snippet.ex
lucacervello: Usually the module in `test/support` don't use `Test` in the module name like https://github.com/elixir-tools/next-ls/blob/main/test/support/utils.ex

> Can you explain what the first element of the accumulator signifies here?

do you mean `false`?

mhanberg: I meant the tuple `{:test, []}`. I'm not sure what the atom is doing

lucacervello: It's used [here](https://github.com/elixir-tools/next-ls/pull/398/files#diff-58f5082ef38a94ce607842d596a0ec1d974224124c9648642c2d46f5435ce750R200) to check if there is `support` under `test`

lucacervello: Oh my bad, I'll try to explain.

I'm looking for three folders

- `lib` where usually the code is, `lib/foo.ex` it's translated to `Foo` (without `lib`)
- `test` where usually the tests are, `test/foo_test.exs` it's translated to `FooTest` (without `test`)
- `support` where usually are utils for tests, `test/support/foo.ex` it's translated to `Foo` (without `test` and `support`)

here are some tests that can help https://github.com/elixir-tools/next-ls/blob/9d65b759ac0cf86c943190342cf829097fc62346/test/next_ls/snippet_test.exs

mhanberg: oh, i understand now, thanks!

basically you are just striping `lib/`, `test/` and `test/support/`  from the beginning for the path.

lucacervello: Yes 👍

lib/next_ls/helpers/ast_helpers/env.ex
```diff
@@ -113,6 +113,8 @@ defmodule NextLS.ASTHelpers.Env do
     }
   end
 
+  def ascend(nil, acc, _callback), do: acc
+
```
mhanberg, line 117: is this change necessary?

mhanberg (approved):
will merge after my last comment is answered.

lib/next_ls/helpers/ast_helpers/env.ex
lucacervello: Yes

without it
```elixir
defmod|
```
don't trigger autocomplete because there is no previous node 

line here: https://github.com/elixir-tools/next-ls/pull/398/files/9d65b759ac0cf86c943190342cf829097fc62346#diff-934a3387a31dbe6cdfbab7475b2f2c24722ff71a7c11b282ef363ac6952f1c9fR19

lib/next_ls/snippet.ex
scottming: @lucacervello When you use someone else's code, please mention the source. If possible, contacting the author would be best; otherwise, it appears quite unethical.

https://github.com/lexical-lsp/lexical/blob/2178b4beab5d1a04cb83990e09b3dc232ed941cb/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex#L567C5-L603C8

mhanberg: I've removed the code on main.

Sorry about that!

PR #408, opened 2024-04-04
fix(completions): correctly insert cursor position

### LINKED ISSUE #406: Completions error when triggered outside of a module
## Description

When typing outside of a module, completions are not showing up.

I believe that this is due to calculating the module name to include in the defmodule snippet.

mhanberg:
The cursor was being inserted in a strange matter, which made the
modified ast not have a `__cursor__()` function, so we couldn't find the
cursor.

Really, all we need to do is insert the cursor on its own line between
the line.

For example, if you were to put the cursor into the middle of this line
of code

```elixir
Enum.map(some_list, fn row ->   end)
```

It would insert the cursor like so

```elixir
Enum.map(some_list, fn row ->
__cursor__()
end)
```

Fixes #406


mhanberg:
* **#408** <a href="https://app.graphite.dev/github/pr/elixir-tools/next-ls/408?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @mhanberg and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

PR #428, opened 2024-04-20
feat: add extract variable command

NJichev:
Searches the AST for an expression of the type:
`variable = expression` where the cursor can be anywhere in the expression.
Moves the variable as a module attribute and any other references of this variable are refactored as well. Afterwards it removes the assignment expression.


NJichev:
Closing for now

PR #441, opened 2024-04-24
undefined function code action

mhanberg:
When you have an undefined local function, this code action allows you
to create a private function stub for the function.


mhanberg:
CleanShot 2024-04-24 at 19 24 13@2x
enhanced with public and private function code action

NJichev:
Looks good, maybe we can add a #TODO: Add code comment in the body

PR #444, opened 2024-04-30
refactor(completions): improve cursor position detection

mhanberg:
Added a spitfire equivalent of the [Code.Fragment.container_cursor_to_quoted](https://hexdocs.pm/elixir/Code.Fragment.html#container_cursor_to_quoted/2) function and that gives us the cursor position.

That will need some improvements, but is currently blocked on improving the lexer.


mhanberg:
way too slow

lib/next_ls/helpers/ast_helpers.ex
```diff
@@ -204,20 +204,89 @@ defmodule NextLS.ASTHelpers do
...
-    else
-      zipper -> {:ok, zipper}
-    end
+  def find_cursor(ast, position) do
+    ast = {:__block__, [], [ast]}
+
+    {tree, id} =
+      ast
+      |> Zipper.zip()
+      |> Zipper.traverse(0, fn tree, id ->
```
mhanberg, line 213: probably want to keep an accumulator of the zipper when its at the spot it inserts the id, then instead of the second pass, we'll already be at the spot on the zipper to inject the cursor

mhanberg:
error raised in Sourceror.Range.get_end_pos_for_interpolation_segments

code located in my notes as it contains work code

lib/next_ls/helpers/ast_helpers/env.ex
```diff
@@ -7,115 +7,119 @@ defmodule NextLS.ASTHelpers.Env do
...
+              _ ->
+                acc
+            end
+          end)
 
-    %{
-      variables: Enum.uniq(env.variables)
-    }
+        %{
+          variables: Enum.uniq(env.variables)
```
NJichev, line 120: Probably all the zipper stuff is slow but you could use MapSet here.

lib/next_ls/helpers/ast_helpers.ex
```diff
@@ -204,19 +204,68 @@ defmodule NextLS.ASTHelpers do
...
+          left && Zipper.node(left) == :do ->
+            Zipper.update(cursor_tree, fn n ->
+              {:__block__, [], [n, {:__cursor__, [], []}]}
+            end)
+
+          true ->
+            Zipper.insert_right(cursor_tree, {:__cursor__, [], []})
+        end
+
+      Zipper.top(cond_result)
```
NJichev, line 266: Maybe you can return cond_result as well, wouldn't that be the zipper being at the cursor which you find in the `with` clause in the build step.

mhanberg:
Turns out most of the slowness comes from fetching the docs for every function and module. 

Will omit those and fetch on the completionItem/resolve request

PR #445, opened 2024-05-03
feat: add folding range provider

### LINKED ISSUE #56: folding

NJichev:
Closes #56 
Adds a folding range provider. For now it only works on do end blocks.
Things to consider:
  - typespecs and docs and whether to fold them together with their respsective function - makes this PR a bit harder and maybe it's not necessary for starters?
  - folding on `fn -> end` blocks - I think this should be added as well in the first version

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -158,7 +158,7 @@ defmodule NextLS do
              "alias-refactor"
            ]
          },
-        folding_range_provider: true,
+         folding_range_provider: true,
          hover_provider: true,
          workspace_symbol_provider: true,
          document_symbol_provider: true,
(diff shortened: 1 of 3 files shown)


mhanberg:
The folds I added to nvim-treesitter currently consist of 

```

[
  (anonymous_function)
  (stab_clause)
  (arguments)
  (block)
  (do_block)
  (list)
  (map)
  (tuple)
] @fold
```

So should at least do those. 

Eventually it would be nice to fold consecutive `alias Foo.Bar`, as we have a lot of them in the Next LS code base, would be cool to fold them up.

NJichev:
Thanks for the feedback, I'll go and implement those, will edit the original comment with a todo list for those.

PR #466, opened 2024-05-10
Copy files over first, then apply patchelf

knarkzel:
Work in progress trying to fix `next-ls` for NixOS.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -134,35 +134,40 @@
           BURRITO_ERTS_PATH = "/tmp/beam/";
           BURRITO_TARGET = lib.optional localBuild burritoExe.${system};
 
-          preBuild =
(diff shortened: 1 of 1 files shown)


knarkzel:
Fixed permission denied issue, now I have this:
```
~/elixir/live_view_studio ❯ ls
_build  config  lib   test     mix.lock  README.md
assets  deps    priv  mix.exs  out
~/elixir/live_view_studio ❯ nextls --stdio=true
Runtime terminating during boot (terminating)

Crash dump is being written to: erl_crash.dump...done
```

knarkzel:
31mb crash dump:
```
.rw-r-----   31M odd  10 May 21:10  erl_crash.dump
```

PR #472, opened 2024-05-13
feat: include `do` as a completions item/snippet

mhanberg:
cc @teamon


teamon:
🔥 🔥 🔥 

I tested that locally in vscode and it works great, thanks!

PR #491, opened 2024-05-20
fix: be more resilient LSP protocol errors

mhanberg:
fix: properly log unknown workspace command


mhanberg:
going to test this branch out at work today before merging.

PR #506, opened 2024-05-31
fix: support numbered lists in docs to prevent crash

fhunleth:
This was triggered by hovering over `:gen_tcp.shutdown/2`.


.tool-versions
mhanberg, line 1: can you remove this change?

PR #510, opened 2024-06-05
fix(runtime): use correct archive locations (e.g. Hex) for bundled Elixir

soundmonster:
Set `MIX_ARCHIVES` explicitly to make sure to override any existing setting, e.g. from `asdf` or `mise`.

What happens on `main` currently is that `hex` is installed into whatever the environment variable `MIX_ARCHIVES` points at. The var is set by e.g. `mise`. So if the "regular" project setup is running on Elixir < 1.17, its installation of Hex will be overridden by NextLS and it stops working after that.

This patch makes sure to set `MIX_ARCHIVES` explicitly so the `hex` archive ends up in the right place.

### DIFF
--- a/lib/next_ls.ex
+++ b/lib/next_ls.ex
@@ -77,7 +77,8 @@ defmodule NextLS do
         :extensions,
         :registry,
         :bundle_base,
-        :mix_home
+        :mix_home,
+        :mix_archives
       ])
 
     GenLSP.start_link(__MODULE__, args, opts)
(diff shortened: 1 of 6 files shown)


soundmonster:
After a closer look I realized that installing into the right folder is only one part of it, we also need to make sure we call Hex from that place going forward. Hence the second commit that sets `MIX_ARCHIVES` in all places where `MIX_HOME` is being set. This makes sure to override what `asdf`/`mise` have set up, so that `hex` works with bundled `mix` even if it's not installed outside of the LSP.

mhanberg:
Thanks!

PR #525, opened 2024-06-24
chore(nix): updates to use nixpkgs elixir 1.17

adamcstephens:
I split the package out here, and dropped the makeOverridable. I can put this back if needed, but I think separating the package makes it clearer to reason about. It's also more similar to a nixpkgs package this way. :)

I discovered that erlang_27 actually depends on erlang_26 currently for ex_doc, so you have to compile both with beam_minimal. Anybody using your cache won't notice though.

I also wondered if the zig pinned release could be removed in favor of current unstable.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,28 +18,18 @@
   }: let
     inherit (nixpkgs) lib;
 
-    version = "0.23.0"; # x-release-please-version
-
     # Helper to provide system-specific attributes
     forAllSystems = f:
       lib.genAttrs systems (system: let
         pkgs = nixpkgs.legacyPackages.${system};
         zpkgs = zigpkgs.legacyPackages.${system};
-        beamPackages = pkgs.beam_minimal.packages.erlang_27;
-        otp = (pkgs.beam.packagesWith beamPackages.erlang).extend (final: prev: {
-          elixir_1_17 = prev.elixir_1_16.override {
-            rev = "v1.17.0";
-            # You can discover this using Trust On First Use by filling in `lib.fakeHash`
-            sha256 = "sha256-RBylCfD+aCsvCqWUIvqXi3izNqqQoNfQNnQiZxz0Igg=";
-            version = "1.17.0";
-          };
-
+        beamPackages = pkgs.beam_minimal.packages.erlang_27.extend (final: prev: {
           elixir = final.elixir_1_17;
           # This will get upstreamed into nix-beam-flakes at some point
           rebar = prev.rebar.overrideAttrs (_old: {doCheck = false;});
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -18,28 +18,18 @@
...
-        beamPackages = pkgs.beam_minimal.packages.erlang_27;
-        otp = (pkgs.beam.packagesWith beamPackages.erlang).extend (final: prev: {
-          elixir_1_17 = prev.elixir_1_16.override {
-            rev = "v1.17.0";
-            # You can discover this using Trust On First Use by filling in `lib.fakeHash`
-            sha256 = "sha256-RBylCfD+aCsvCqWUIvqXi3izNqqQoNfQNnQiZxz0Igg=";
-            version = "1.17.0";
-          };
-
+        beamPackages = pkgs.beam_minimal.packages.erlang_27.extend (final: prev: {
```
mhanberg, line 26: i don't think we need to extend this anymore

mhanberg:
> I also wondered if the zig pinned release could be removed in favor of current unstable.

something is broken in nixpkgs for zig 0.11 right now, which we need, so we can't do that yet.

---

You'll halso have to change the `extra_files` entry in `release-please-config.json` to point to the package.nix file instead of flake.nix file

flake.nix
adamcstephens: I left this in because two erlangs have to be compiled for this config, both 26 and 27. https://github.com/NixOS/nixpkgs/issues/322168

I'm happy to remove it, but anybody building will have to wait on the slow rebar tests to run.

mhanberg: I think that happens anyway, when i pulled down the branch and my direnv starting booting up the nix develop shell, it compiled otp 26 and rebar

adamcstephens: Yeah we can't avoid compiling for now, but running the *tests* for rebar add significant time to the build. I'll pull it out since most users will be using the cache anyway.

mhanberg: Does extending it remove the tests?

adamcstephens: Yes. It's effectively an override for these nested packages allowing the tests to be disabled.

mhanberg: oh, is that what the `doCheck = false` is doing? I had copied that from an example previously and didn't pay it much attention.

adamcstephens: Yes :)

mhanberg (approved):
Just make the change to the release please config and looks good, thanks!

adamcstephens:
I think this should be good to go, unless you still wanted me to remove the rebar overrides.

mhanberg:
I checked out the branch locally and as my shell was booting up nix develop, its running the the rebar tests.

still better than before, will merge once i get a `nix build` passing locally

